PR #1, opened 2024-09-26
feat(redis): implement redis caches

hannahhoward:
# Goal

Implement the three types of cache interfaces based on the redis cache

# Implementation

- built a general purpose redis cache interface talking to default go redis client library
- built three special cases to handle serialization/deserialization for ipni cache, the sharded dag index cache, and the content claims cache
- also fixed a few issues with assert & metadata cause the tests finally exercised them a bit
- added a bunch of test infra (a lot copied over from go-ucanto)


pkg/metadata/metadata.ipldsch
```diff
@@ -0,0 +1,5 @@
+type ContentClaimMetadata struct {
+  ClaimType Int
+  ShortCut Bytes
```
alanshaw, line 3: 🙏 can this be named better? Something like "embedded data"/"cached data"/"claimed data"?

pkg/metadata/metadata.ipldsch
```diff
@@ -0,0 +1,5 @@
...
+  ClaimCID &Any
```
alanshaw, line 4: Should be a `Link` type?

alanshaw: IDK personally I don't like to have type names in field names...

```suggestion
  Claim Link
```

pkg/redis/contentclaimsstore.go
```diff
@@ -0,0 +1,38 @@
...
+// NewContentClaimsStore returns a new instance of a Content Claims Store using the given redis client
+func NewContentClaimsStore(client Client) *ContentClaimsStore {
+	return &Store[cid.Cid, delegation.Delegation]{delegationFromRedis, delegationToRedis, cidKeyString, client}
+}
+
+func delegationFromRedis(data string) (delegation.Delegation, error) {
+	return delegation.Extract([]byte(data))
+}
+
+func delegationToRedis(d delegation.Delegation) (string, error) {
```
alanshaw, line 27: You can store bytes in redis no?

pkg/redis/providerresult.ipldsch
```diff
@@ -0,0 +1,18 @@
+type PeerID bytes
+type Multiaddr bytes
```
alanshaw, line 2:
```suggestion
type PeerID Bytes
type Multiaddr Bytes
```

pkg/redis/redisstore.go
```diff
@@ -0,0 +1,89 @@
...
+		if err == redis.Nil {
+			return v, types.ErrKeyNotFound
+		}
+		return v, fmt.Errorf("error accessing redis: %w", err)
+	}
+	return rs.fromRedis(data)
+}
+
+// Set saves a serialized value to redis
+func (rs *Store[Key, Value]) Set(ctx context.Context, key Key, value Value, expires bool) error {
```
alanshaw, line 61: What's the reason to use a boolean vs a duration for expiration?

pkg/redis/redisstore.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	err = rs.client.Set(ctx, rs.keyString(key), data, duration).Err()
+	if err != nil {
+		return fmt.Errorf("error accessing redis: %w", err)
+	}
+	return nil
+}
+
+// SetExpirable changes the expiration property for a given key
+func (rs *Store[Key, Value]) SetExpirable(ctx context.Context, key Key, expires bool) error {
```
alanshaw, line 78: When do we want something that previously had no expiry to have one (and vice versa)?

pkg/redis/redisstore_test.go
```diff
@@ -0,0 +1,187 @@
...
+		})
+	}
+}
+
+type redisValue struct {
+	data    string
+	expires time.Duration
+}
+
+type MockRedis struct {
```
alanshaw, line 95: There's a mock redis implementation here https://github.com/go-redis/redismock - is this not suitable?

alanshaw (changes requested):
🚀 this looks pretty good to me! Just some minor feedback - mostly questions :)

pkg/capability/assert/assert.go
```diff
@@ -44,24 +44,22 @@ func Link(l datamodel.Link) (HasMultihash, failure.Failure) {
 	return link{l}, nil
 }
 
-type digest struct {
-	Digest mh.Multihash
-}
+type digest mh.Multihash
 
 func (d digest) hasMultihash() {}
```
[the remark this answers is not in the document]
hannahhoward: ha yea actually was just messing around but it's simpler.

pkg/internal/testutil/helpers.go
```diff
@@ -0,0 +1,56 @@
...
+// the error value is non-nil then it fails the test.
+func Must2[T, U any](val1 T, val2 U, err error) func(*testing.T) (T, U) {
+	return func(t *testing.T) (T, U) {
+		require.NoError(t, err)
+		return val1, val2
+	}
+}
+
+// RequireEqualIndex compares two sharded dag indexes to verify their equality
+func RequireEqualIndex(t *testing.T, expectedIndex blobindex.ShardedDagIndexView, actualIndex blobindex.ShardedDagIndexView) {
```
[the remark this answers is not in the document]
hannahhoward: yea this is actually used to check a archive/unarchive operation so it seems redundant at this point.

pkg/metadata/metadata.ipldsch
hannahhoward: fair

pkg/redis/providerresult.ipldsch
hannahhoward: that actually didn't work. I think when you define alias types you have to use the lower case.

pkg/redis/contentclaimsstore.go
hannahhoward: yes, but everything is a string to anyway.

pkg/redis/redisstore.go
hannahhoward: I dunno simplicity given it's an abstract interface. also the key distinction I'm thinking of is keys you can collect to keep the cache size down vs keys you dont want to do that on.

pkg/redis/redisstore.go
hannahhoward: it's actually the other way around -- I cached on write before IPNI, so I set NO expiry. Now i want to set YES expiry cause I've confirmed it's actually on IPNI now.

pkg/redis/redisstore_test.go
hannahhoward: yea, I saw this -- generally, I want something a bit more stateful, that's simple and which I understand. I often that particular style of mocking library (seems analogous to this https://github.com/uber-go/mock) quite frustrating and awkward to use, and ultimately it doesn't tell you much. By and large, I just write my own mocks in go, cause I want to understand exactly the behavior that's happening under the hood and to see what comes out of that.

hannahhoward: there's also this -- https://github.com/alicebob/miniredis

ultimately I looked at both and was like "it would be easiest for me to write tests against something I expected to behave a very specific way.

alanshaw (approved):

PR #18, opened 2024-10-11
Add space did to location commitment

hannahhoward:
# Goal

Adds an optional space did parameter to the location commitment

# Implementation

Add to assert.model & the IPLD schema, add DID decode/encode on serialization and deserialization.

### DIFF
--- a/pkg/capability/assert/assert.go
+++ b/pkg/capability/assert/assert.go
@@ -11,6 +11,7 @@ import (
 	"github.com/storacha/go-ucanto/core/ipld"
 	"github.com/storacha/go-ucanto/core/result/failure"
 	"github.com/storacha/go-ucanto/core/schema"
+	"github.com/storacha/go-ucanto/did"
 	"github.com/storacha/go-ucanto/validator"
 	adm "github.com/storacha/indexing-service/pkg/capability/assert/datamodel"
 )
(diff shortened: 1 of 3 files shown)


alanshaw (approved):
LGTM.

I initially had concerns over this being part of a location commitment...and was going to suggest it was a separate claim made by the authority of the space, but on balance in terms of garbage collection I think it's useful to know the space you're storing something on behalf of. i.e. if multiple users store the same blob and it happens to be stored on the same node then the location commitments are the reference counts.

I'm just wondering how the storage node comes to know of the space. I'm guessing something like this. Where this echos very closely what we already have, except adding in the third party: the storage provider.

```mermaid
sequenceDiagram
    participant alice
    participant service
    participant storagenode
    alice->>service: blob/add with: alicespace nb: blob=car size=3
    service->>storagenode: blob/allocate with: storagenode nb: space=alicespace blob=car size=3
    storagenode->>service: receipt ran: blob/allocate out: url
    service->>alice: receipt ran: blob/add fx: blob/allocate, http/put, blob/accept <br> (included in response: receipt ran: blob/allocate out: url)
    alice->>storagenode: HTTP PUT
    alice->>service: receipt ran: http/put
    service->>storagenode: blob/accept with: storagenode nb: space=alicespace blob=car
    storagenode->>service: receipt ran: blob/accept out: locationcommitment
    alice->>service: poll to get receipt for blob/accept
```

So, the storage node knows about the space at both `blob/allocate` and `blob/accept`, so will definitely be able to add the space DID to the location commitment.

PR #21, opened 2024-10-14
refactor: remove repeated `Service` interface from server pkg

alanshaw:
Removes repeated `Service` interface from `pkg/server/server.go` and adds to `pkg/types/types.go`. The `Service` defined in `pkg/construct/construct.go` (with lifecycle methods) now extends `types.Service`. Had to move some query structs into `types` to resolve import cycles.

This also now allows the indexing service to be passed as a dependency to the content claims service, and is done here also (although currently unused).

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -132,19 +132,18 @@ func main() {
 							}
 							sc.PrivateKey = privKey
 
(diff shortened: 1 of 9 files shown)


hannahhoward (approved):
Lovely!

PR #22, opened 2024-10-14
feat: publish index claim

alanshaw:
* Wires up the Ucanto endpoints to publish claims when invoked
* Adds implementation for publishing index claims

TODO: add tests


pkg/service/contentclaims/service.go
```diff
@@ -17,22 +19,22 @@ func NewService(indexer types.Service) map[ucan.Ability]server.ServiceMethod[ass
...
+				err := indexer.PublishClaim(context.TODO(), inv)
+				return assert.Unit{}, nil, err
 			},
 		),
 		assert.Location.Can(): server.Provide(
 			assert.Location,
 			func(cap ucan.Capability[assert.LocationCaveats], inv invocation.Invocation, ctx server.InvocationContext) (assert.Unit, receipt.Effects, error) {
-				log.Errorf("TODO: implement me")
-				return assert.Unit{}, nil, nil
+				err := indexer.PublishClaim(context.TODO(), inv)
```
hannahhoward, line 36: This would be for when we publish our own claims for our own service right?

Cause seperately we have the "CacheClaim" call just for the location commitments manged by others.

hannahhoward: sorry rewind looks like neither this nor the equals claim are implemented?

pkg/service/service.go
```diff
@@ -333,3 +345,169 @@ func NewIndexingService(blobIndexLookup BlobIndexLookup, claimLookup ClaimLookup
...
+	var ferr error
+	for _, r := range results {
+		idx, ferr = fetchBlobIndex(ctx, blobIndex, claimLookup, r)
+		if ferr != nil {
+			continue
+		}
+		break
+	}
+	if ferr != nil {
+		return fmt.Errorf("fetching blob index: %w", err)
```
hannahhoward, line 390: it seems like if we had a successful fetch,  even if another errored, then we can proceed, no?

hannahhoward (commented):
General LGTM. Let's get test and maybe a bit more commentary on logic.

hannahhoward:
also you're failing tests at the moment.

pkg/service/service.go
alanshaw: Oops! Good catch.

alanshaw: No wait, a successul fetch will set `ferr` to `nil`, so we will proceed.

hannahhoward (approved):
LGTM

PR #25, opened 2024-10-17
feat(deploy): add terraform configs

hannahhoward:
# Goals

Enable a full deployment of the indexing service, including with similar deploy setup to w3infra where you can deploy a branched setup based on your user name or PR #

# Implementation

All the CFGs, basically taken from various iterations of different terraform projects I've worked on before.

Note that we haven't quite yet got everything -- still need to setup a number of access policies and roles.


alanshaw (commented):
🙏 Please can we ensure that all infra created/managed is tagged with a tag called "Project" so we can attribute costs appropriately.

hannahhoward:
in the interest of keeping everything on track, merging

PR #29, opened 2024-10-24
feat: cache and store content claims

alanshaw:
This PR changes the `ClaimLookup` interface to allow caching and permanent storage (publishing) of content claims. It also allows retrieval by content claim root CID.


pkg/blobindex/shardeddagindex.go
```diff
@@ -77,7 +77,7 @@ func View(root ipld.Link, blockMap map[ipld.Link]ipld.Block) (ShardedDagIndexVie
 	if shardedDagIndexData.DagO_1 == nil {
 		return nil, NewUnknownFormatError(fmt.Errorf("unknown index version"))
 	}
-	dagIndex := NewShardedDagIndexView(root, len(shardedDagIndexData.DagO_1.Shards))
+	dagIndex := NewShardedDagIndexView(shardedDagIndexData.DagO_1.Content, len(shardedDagIndexData.DagO_1.Shards))
```
alanshaw, line 80: Bug where a decoded blob index content field was being set to the CID of the CAR.

pkg/construct/construct.go
```diff
@@ -255,6 +255,7 @@ func Construct(sc ServiceConfig, opts ...Option) (Service, error) {
 
 		s.startupFuncs = append(s.startupFuncs, func(context.Context) error { jq.Startup(); return nil })
 		s.shutdownFuncs = append(s.shutdownFuncs, jq.Shutdown)
+		cachingQueue = jq
```
alanshaw, line 258: Bug `cachingQueue` remained `nil`.

pkg/service/service.go
```diff
@@ -282,7 +283,7 @@ func (is *IndexingService) Get(ctx context.Context, claim ipld.Link) (delegation
 // ideally however, IPNI would enable UCAN chains for publishing so that we could publish it directly from the storage service
 // it doesn't for now, so we let SPs publish themselves them direct cache with us
 func (is *IndexingService) Cache(ctx context.Context, provider peer.AddrInfo, claim delegation.Delegation) error {
-	return Cache(ctx, is.blobIndexLookup, is.claims, is.providerIndex, is.provider, claim)
+	return Cache(ctx, is.blobIndexLookup, is.claims, is.providerIndex, provider, claim)
```
alanshaw, line 286: Bug - we were using the indexer provider info when caching claims from storage nodes 🤦

pkg/metadata/metadata.go
```diff
@@ -99,6 +92,10 @@ func init() {
 	mdctx = mdctx.WithProtocol(EqualsClaimID, func() ipnimd.Protocol { return &EqualsClaimMetadata{} })
 	mdctx = mdctx.WithProtocol(LocationCommitmentID, func() ipnimd.Protocol { return &LocationCommitmentMetadata{} })
 	MetadataContext = mdctx
+	// WithProtocol creates a _new_ context. Assign it to Default so claim
+	// metadata can be decoded elsewhere by refrencing libipni metadata.Default
+	// i.e. ipni-publisher
+	ipnimd.Default = mdctx
```
alanshaw, line 98: Bug - see comment. The `ipni-publisher` store uses `metadata.Default` and was not able to decode our custom metadatas. I think ideally we'd pass metadata context through?

pkg/metadata/metadata.go
```diff
@@ -85,12 +84,6 @@ const EqualsClaimID = 0x3E0001
 // LocationCommitmentID is the multicodec for location commitments
 const LocationCommitmentID = 0x3E0002
 
-var nodePrototypes = map[multicodec.Code]schema.TypedPrototype{
-	IndexClaimID:         indexClaimMetadata,
-	EqualsClaimID:        equalsClaimMetadata,
-	LocationCommitmentID: locationCommitmentMetadata,
-}
```
alanshaw, line 92: Bug - These were `nil` as `init` has not run yet.

pkg/construct/construct.go
hannahhoward: #oops

deploy/s3.tf
```diff
@@ -47,4 +47,8 @@ resource "aws_s3_bucket_policy" "ipni_store_policy" {
 
 resource "aws_s3_bucket" "notifier_head_bucket" {
   bucket = "${terraform.workspace}-${var.app}-notifier-head-bucket"
-}
\ No newline at end of file
+}
+
+resource "aws_s3_bucket" "claim_store_bucket" {
```
hannahhoward, line 52: Question: in prod, shouldn't this be the same bucket we already have to legacy claims? Or do we store it different.

pkg/metadata/metadata.go
hannahhoward: yea we should

pkg/service/service.go
```diff
@@ -28,62 +29,28 @@ import (
...
-	// other multihashes in the index
-	Find(ctx context.Context, contextID types.EncodedContextID, provider model.ProviderResult, fetchURL url.URL, rng *metadata.Range) (blobindex.ShardedDagIndexView, error)
-}
-
 // IndexingService implements read/write logic for indexing data with IPNI, content claims, sharded dag indexes, and a cache layer
 type IndexingService struct {
-	blobIndexLookup BlobIndexLookup
-	claimLookup     ClaimLookup
-	providerIndex   ProviderIndex
+	blobIndexLookup blobindexlookup.BlobIndexLookup
```
hannahhoward, line 44: Apologies for my GO related thinking here -- there's a meme in go of declare the interface where you use it. Honestly, not sure I've ever fully bought in :)

pkg/service/service.go
hannahhoward: Fle under: oops.

hannahhoward (changes requested):
Generally awesome and LGTM

One big caveat, which I realize is actually my own mistake. I extracted the use of metadata to the go-metadata package but failed to delete the actual metadata file. OOPS -- we should do that right away to avoid confusion

alanshaw:
> One big caveat, which I realize is actually my own mistake. I extracted the use of metadata to the go-metadata package but failed to delete the actual metadata file. OOPS -- we should do that right away to avoid confusion

Ah, damn, yes ok that explains a lot. I thought I'd seen those issues before 🤦

hannahhoward (approved):

pkg/service/service.go
alanshaw: i_have_no_idea_what_im_doing

PR #31, opened 2024-10-25
feat(client): client package for indexer

hannahhoward:
# Goals

A golang client to talk to the indexer

# Implementation

1. Client package
2. Go ucanto minor update

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -12,6 +12,7 @@ import (
 	"github.com/storacha/go-capabilities/pkg/assert"
 	"github.com/storacha/go-capabilities/pkg/claim"
 	"github.com/storacha/go-ucanto/client"
+	"github.com/storacha/go-ucanto/core/delegation"
 	"github.com/storacha/go-ucanto/core/invocation"
 	"github.com/storacha/go-ucanto/core/receipt"
 	"github.com/storacha/go-ucanto/core/result"
@@ -35,7 +36,7 @@ type ErrFailedResponse struct {
 	Body       string
 }
 
-func errFromResponse(res *http.Response) ErrFailedResponse {
+func errFromResponse(res *gohttp.Response) ErrFailedResponse {
 	err := ErrFailedResponse{StatusCode: res.StatusCode}
 
 	message, merr := io.ReadAll(res.Body)
@@ -48,29 +49,17 @@ func errFromResponse(res *http.Response) ErrFailedResponse {
 }
 
 func (e ErrFailedResponse) Error() string {
-	return fmt.Sprintf("http request failed, status: %d %s, message: %s", e.StatusCode, http.StatusText(e.StatusCode), e.Body)
+	return fmt.Sprintf("http request failed, status: %d %s, message: %s", e.StatusCode, gohttp.StatusText(e.StatusCode), e.Body)
 }
 
 type Client struct {
 	servicePrincipal ucan.Principal
-	serviceURL       string
-}
-
-func (c *Client) connect() (client.Connection, error) {
-	url, err := url.Parse(c.serviceURL)
-	if err != nil {
-		return nil, fmt.Errorf("parsing service URL: %w", err)
-	}
-	return client.NewConnection(c.servicePrincipal, http.NewHTTPChannel(url.JoinPath(claimsPath)))
+	serviceURL       url.URL
+	connection       client.Connection
 }
 
 func (c *Client) execute(inv invocation.Invocation) error {
-	connection, err := c.connect()
-	if err != nil {
-		return fmt.Errorf("establishing client connection: %w", err)
-	}
-
-	resp, err := client.Execute([]invocation.Invocation{inv}, connection)
+	resp, err := client.Execute([]invocation.Invocation{inv}, c.connection)
 	if err != nil {
 		return fmt.Errorf("sending invocation: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return fmt.Errorf("reading receipt: %w")
+	}
+
+	_, err = result.Unwrap(result.MapError(rcpt.Out(), failure.FromFailureModel))
+	return err
+}
+
+func (c *Client) PublishIndexClaim(ctx context.Context, issuer principal.Signer, caveats assert.IndexCaveats) error {
+	inv, err := assert.Index.Invoke(issuer, c.servicePrincipal, c.servicePrincipal.DID().String(), caveats)
```
alanshaw, line 97: Need to be able to provide proofs

pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return fmt.Errorf("generating invocation")
+	}
+	return c.execute(inv)
+}
+
+func (c *Client) PublishEqualsClaim(ctx context.Context, issuer principal.Signer, caveats assert.EqualsCaveats) error {
+	inv, err := assert.Equals.Invoke(issuer, c.servicePrincipal, c.servicePrincipal.DID().String(), caveats)
```
alanshaw, line 105: Again proofs

pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+		err.Body = string(message)
+	}
+	return err
+}
+
+func (e ErrFailedResponse) Error() string {
+	return fmt.Sprintf("http request failed, status: %d %s, message: %s", e.StatusCode, http.StatusText(e.StatusCode), e.Body)
+}
+
+type Client struct {
```
alanshaw, line 54: Need a constructor

pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	servicePrincipal ucan.Principal
+	serviceURL       string
+}
+
+func (c *Client) connect() (client.Connection, error) {
+	url, err := url.Parse(c.serviceURL)
+	if err != nil {
+		return nil, fmt.Errorf("parsing service URL: %w", err)
+	}
+	return client.NewConnection(c.servicePrincipal, http.NewHTTPChannel(url.JoinPath(claimsPath)))
```
alanshaw, line 64: Connections can be re-used, don't have to create for every request.

pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	_, err = result.Unwrap(result.MapError(rcpt.Out(), failure.FromFailureModel))
+	return err
+}
+
+func (c *Client) PublishIndexClaim(ctx context.Context, issuer principal.Signer, caveats assert.IndexCaveats) error {
+	inv, err := assert.Index.Invoke(issuer, c.servicePrincipal, c.servicePrincipal.DID().String(), caveats)
+	if err != nil {
+		return fmt.Errorf("generating invocation")
```
alanshaw, line 99:
```suggestion
		return fmt.Errorf("generating invocation: %w", err)
```

pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return c.execute(inv)
+}
+
+func (c *Client) PublishEqualsClaim(ctx context.Context, issuer principal.Signer, caveats assert.EqualsCaveats) error {
+	inv, err := assert.Equals.Invoke(issuer, c.servicePrincipal, c.servicePrincipal.DID().String(), caveats)
+	if err != nil {
+		return fmt.Errorf("generating invocation")
```
alanshaw, line 107:
```suggestion
		return fmt.Errorf("generating invocation: %w", err)
```

pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return c.execute(inv)
+}
+
+func (c *Client) CacheClaim(ctx context.Context, issuer principal.Signer, provider claim.Provider, caveats assert.LocationCaveats) error {
```
alanshaw, line 112: Need proofs again...

pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	lc, err := assert.Location.Invoke(issuer, issuer.DID(), caveats.Space.String(), caveats)
+	if err != nil {
+		return fmt.Errorf("building location commitment: %w", err)
+	}
+	inv, err := claim.Cache.Invoke(issuer, c.servicePrincipal, c.servicePrincipal.DID().String(), claim.CacheCaveats{
+		Claim:    lc.Link(),
+		Provider: provider,
+	})
```
alanshaw, line 120: What if we just pass a `provider` and `delegation.Delegation` to the function? It can set the claim link from the delegation and `Attach` it to the invocation. We need the location commitment to be created outside of this function as it need to be returned in the `blob/accept` receipt.

pkg/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("encoding query multihash")
+		}
+		q.Add("multihash", mhString)
+	}
+	for _, space := range query.Match.Subject {
+		q.Add("spaces", space.String())
+	}
+	url.RawQuery = q.Encode()
+	res, err := gohttp.DefaultClient.Get(url.String())
```
alanshaw, line 154: Should probably allow this to be configured...

alanshaw (changes requested):

alanshaw:
FYI - I have fixed this up and merged into https://github.com/storacha/indexing-service/pull/29

PR #34, opened 2024-11-05
feat: tool to generate a query result test fixture

alanshaw:
Tool to generate a test fixture allowing a JS (or other language) client to be tested.

Moves bytemap out of internal since `queryresult.Build(..)` requires it.


volmedo (approved):
LGTM

Since it's a new command, I'd place it under `cmd` along with the others. It could also be interesting to add a new target to the `Makefile` to run the tool.

Feel free to merge as is. If you think the suggestions above make sense, let me know and I'll move things around afterwards so that you don't need to spend time in that now.

I'm also curious why we use both `pkg` and `internal`. It's likely we only need `internal` for this repo, which would've avoided the need of moving `bytemap` out. But that's something we can discuss at some other time.

alanshaw:
@volmedo I'm not 100% sure about moving to `cmd`, we are following https://github.com/golang-standards/project-layout and the advice is to only put main applications in `cmd`, with `tools` being for supporting...tools 🙄.

volmedo:
that's totally fine with me.

we might want to move `cmd/ucangen` to `tools` then :P

PR #36, opened 2024-11-08
fix: makefile

alanshaw:
Fixes the makefile so a developer can get started _without_ THERE BE DRAGONS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ clean-terraform:
 
 .PHONY: clean
 
-clean: clean-lambda clean-terraform
+clean: clean-terraform clean-lambda
 
 lambdas: $(LAMBDAS)
 
--- a/Makefile
+++ b/Makefile
@@ -4,12 +4,12 @@ ifneq (,$(wildcard ./.env))
 else
   $(error You haven't setup your .env file. Please refer to the readme)
 endif
-GOOS=linux
-GOARCH=arm64
-GOCC?=go
-GOFLAGS=-tags=lambda.norpc
-CGO_ENABLED=0
-LAMBDAS=build/getclaims/bootstrap build/getroot/bootstrap build/notifier/bootstrap build/postclaims/bootstrap build/providercache/bootstrap build/remotesync/bootstrap
+LAMBDA_GOOS=linux
+LAMBDA_GOARCH=arm64
+LAMBDA_GOCC?=go
+LAMBDA_GOFLAGS=-tags=lambda.norpc
+LAMBDA_CGO_ENABLED=0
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -4,12 +4,12 @@ ifneq (,$(wildcard ./.env))
...
-GOARCH=arm64
-GOCC?=go
-GOFLAGS=-tags=lambda.norpc
-CGO_ENABLED=0
-LAMBDAS=build/getclaims/bootstrap build/getroot/bootstrap build/notifier/bootstrap build/postclaims/bootstrap build/providercache/bootstrap build/remotesync/bootstrap
+LAMBDA_GOOS=linux
+LAMBDA_GOARCH=arm64
+LAMBDA_GOCC?=go
+LAMBDA_GOFLAGS=-tags=lambda.norpc
+LAMBDA_CGO_ENABLED=0
```
alanshaw, line 11: Overwriting these env vars causes problems when building for local execution i.e. `ucankey`

Makefile
```diff
@@ -4,12 +4,12 @@ ifneq (,$(wildcard ./.env))
...
+LAMBDAS=build/getclaim/bootstrap build/getclaims/bootstrap build/getroot/bootstrap build/notifier/bootstrap build/postclaims/bootstrap build/providercache/bootstrap build/remotesync/bootstrap
```
alanshaw, line 12: Added missing `getclaim` lambda.

Makefile
```diff
@@ -38,10 +38,10 @@ lambdas: $(LAMBDAS)
 .PHONY: $(LAMBDAS)
 
 $(LAMBDAS): build/%/bootstrap:
-	GOOS=$(GOOS) GOARCH=$(GOARCH) CGO_ENABLED=$(CGO_ENABLED) $(GOCC) build $(GOFLAGS) -o $@ cmd/lambda/$*/main.go
+	GOOS=$(LAMBDA_GOOS) GOARCH=$(LAMBDA_GOARCH) CGO_ENABLED=$(LAMBDA_CGO_ENABLED) $(LAMBDA_GOCC) build $(LAMBDA_GOFLAGS) -o $@ cmd/lambda/$*/main.go
 
 deploy/.terraform:
-	tofu -chdir=deploy init
+	TF_WORKSPACE= tofu -chdir=deploy init
```
alanshaw, line 44: Ok so, tofu expects no existing `TF_WORKSPACE` when initializing, but then needs it subsequently.

Makefile
volmedo: yeah, that makes sense as they are the vars Go uses

volmedo (approved):
LGTM!

hannahhoward (commented):
LGTM

PR #38, opened 2024-11-08
feat(deploy): seperate shared state

hannahhoward:
# Goals

This fixes a bug that comes up with multiple workspaces around the hosted zone so that lives in a seperate workspace. It also adds CI for deploys

# Implementation

- split configs into `shared` (everyone uses these deploy cvonfigs) and `app` -- where most configs live that get deployed on a workspace specific context


deploy/app/lambda.tf
```diff
@@ -68,7 +68,7 @@ resource "aws_lambda_function" "lambda" {
         NOTIFIER_HEAD_BUCKET_NAME = aws_s3_bucket.notifier_head_bucket.bucket
         NOTIFIER_SNS_TOPIC_ARN = aws_sns_topic.published_advertisememt_head_change.id
         PRIVATE_KEY = aws_ssm_parameter.private_key.name
-        PUBLIC_URL = var.public_url
+        PUBLIC_URL = aws_apigatewayv2_domain_name.custom_domain.domain_name
```
alanshaw, line 71: yes, nice I was going to do this...

alanshaw (approved):
🚀 LGTM

PR #52, opened 2024-12-06
feat: legacy claim materialize

alanshaw:
This is a port of the code from https://github.com/storacha/content-claims/blob/bffff9d9c7398daa8c197b2b98af2dd9eee33de3/packages/infra/src/lib/store/block-index.js#L38-L85

The possibly contentious issue is that I'm using testcontainers in the AWS tests to spin up a real life dynamodb instance. Unfortunately testcontainers do not work in mac/win (I didn't realise until after), but given we don't deploy to these environments I don't think it's really an issue. The one dependency it introduces is thta you need docker running on your local dev system to run ALL the tests now.

TODO
* [x] update AWS lambda code to make use of the legacy claim materialization.


pkg/aws/dynamoblockindextable.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	keyEx := expression.Key("multihash").Equal(expression.Value(digestAttr))
+	proj := expression.NamesList(
+		expression.Name("carpath"),
+		expression.Name("offset"),
+		expression.Name("length"),
+	)
```
volmedo, line 39: are there any other fields? If not, you can get rid of the projection

pkg/service/legacy/service.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(results.Claims()) > 0 || len(results.Indexes()) > 0 || len(q.Hashes) == 0 {
+		return results, nil
+	}
+
+	// lets see if we can materialize some location claims
+	content := assert.FromHash(q.Hashes[0])
```
volmedo, line 59: we could get more than a single hash in the query, couldn't we?

pkg/service/legacy/service.go
```diff
@@ -0,0 +1,158 @@
...
+			Content:  content,
+			Location: []url.URL{*u},
+			Range:    &assert.Range{Offset: r.Offset, Length: &r.Length},
+		})
+	}
+
+	// prefer items with non derived location URLs
+	if len(locs) == 0 {
+		locs = derivedLocs
+	}
```
volmedo, line 98: not sure I get why we need `locs` and `derivedLocs`. `records` are data about blocks associated with some `content`. Can these records have mixed `CarPath` formats? I mean, I'd expect all records to either have `CarPath`s that are URLs or legacy. If they all have the same format, we will have all items in either `locs` or `derivedLocs` and we only need one of them.

From the code, I assume that the block index store may contain data for all the blocks that belong to `content` in both formats, is that correct?

volmedo (commented):
looking good so far.

I'm taking some ideas for my PR :)

pkg/service/legacy/service.go
alanshaw: It depends, if the same block is was uploaded in a CAR a long time again and then again in a CAR more recently it'll have multiple entries in the table. That said, I think you're righ it doesn't really matter which ones we use - URLs or old paths... so I've removed.

pkg/aws/dynamoblockindextable_test.go
```diff
@@ -0,0 +1,193 @@
...
+	ctx := context.Background()
+	req := testcontainers.ContainerRequest{
+		Image:        "amazon/dynamodb-local:latest",
+		ExposedPorts: []string{"8000/tcp"},
+		WaitingFor:   wait.ForExposedPort(),
+	}
+	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: req,
+		Started:          true,
+	})
```
volmedo, line 135: I took a look at `testcontainers` docs and saw there is a [module for DynamoDB](https://golang.testcontainers.org/modules/dynamodb/). It looks like it makes the syntax a bit more straightforward. I'm not sure it has other advantages over `GenericContainer` though.

pkg/service/legacy/service.go
```diff
@@ -156,3 +152,7 @@ func bucketKeyToShardLink(key string) (ipld.Link, error) {
 	}
 	return cidlink.Link{Cid: shard}, nil
 }
+
+func isAbsoluteURL(u url.URL) bool {
+	return u.Scheme != "" && u.Host != ""
+}
```
volmedo, line 175: there is a `func (*url.URL) IsAbs() bool` function in the standard library, would that work for your use case? Interestingly, it's implementation only checks that `Scheme` is not empty.

pkg/service/legacy/service_test.go
```diff
@@ -0,0 +1,252 @@
...
+		require.True(t, errors.Is(err, errNotImplemented))
+	})
+}
+
+var errNotImplemented = errors.New("not implemented")
+
+type mockIndexingService struct {
+	queryResult types.QueryResult
+	queryError  error
+}
```
volmedo, line 214: alternatively, you could get testify-compatible mocks created automatically using `mockery` and express expectations in tests directly. Take a look at the tests in my PR #51 or ping me if you feel like going that route.

pkg/aws/dynamoblockindextable_test.go
```diff
@@ -0,0 +1,193 @@
...
+	testcontainers.CleanupContainer(t, container)
+	require.NoError(t, err)
+
+	endpoint, err := container.Endpoint(ctx, "")
+	require.NoError(t, err)
+
+	return *testutil.Must(url.Parse("http://" + endpoint))(t)
```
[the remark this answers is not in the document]
alanshaw: Yeah I see...I'm just keen to communicating the fact that this isn't something that's going to be mutated.

pkg/aws/dynamoblockindextable_test.go
alanshaw: Saves a few lines! Thanks!

pkg/service/legacy/service.go
alanshaw: I think that'll do.

pkg/service/legacy/service_test.go
alanshaw: These are so simple I feel like it might be overkill but maybe you could show me how to generate?

pkg/aws/dynamoblockindextable_test.go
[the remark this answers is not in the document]
volmedo: yes, I'm used to that too in other languages. The convention in Go is subtly different, I think because of how pointers are used in pointer receivers in functions to define the type's API. It feels a bit more idiomatic, but feel free to use whatever you prefer.

pkg/service/legacy/service_test.go
volmedo: let's take a look at it in the next demo session, as you suggested 👍🏻

volmedo (approved):
LGTM! 🚀

Issue #53, opened 2024-12-11
enable access to `content-claims` bucket


volmedo:
`indexing-service` needs to access `content-claims` bucket to fetch legacy claims. There are two ways of achieving this:
- enable public access in `content-claims` bucket
- make `indexing-service` use the S3 API to fetch objects as it does with its own buckets

volmedo:
fixed by Alan in https://github.com/storacha/content-claims/pull/82 and https://github.com/storacha/content-claims/pull/83

PR #56, opened 2024-12-12
chore: allow access to legacy buckets and tables

volmedo:
ref #53 

Add the required policies to allow `indexing-service` access legacy resources. Namely:
- `content-claims` DynamoDB table and S3 bucket
- block index table

### DIFF
--- a/deploy/app/legacyclaims.tf
+++ b/deploy/app/legacyclaims.tf
@@ -13,7 +13,11 @@ variable "legacy_claims_bucket_name" {
 variable "legacy_block_index_table_name" {
   description = "The name of the legacy block index DynamoDB table"
   type = string
-  default = "ep-v1-blocks-cars-position"
+  default = ""
+}
+
+locals {
+    inferred_legacy_block_index_table_name = var.legacy_block_index_table_name != "" ? var.legacy_block_index_table_name : "${terraform.workspace == "prod" ? "prod" : "staging"}-ep-v1-blocks-cars-position"
 }
 
 data "aws_s3_bucket" "legacy_claims_bucket" {
@@ -25,7 +29,7 @@ data "aws_dynamodb_table" "legacy_claims_table" {
 }
 
 data "aws_dynamodb_table" "legacy_block_index_table" {
-  name = "${terraform.workspace == "prod" ? "prod" : "staging"}-${var.legacy_block_index_table_name}"
+  name = local.inferred_legacy_block_index_table_name
 }
 
 data "aws_iam_policy_document" "lambda_legacy_dynamodb_query_document" {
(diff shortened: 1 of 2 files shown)


deploy/app/legacyclaims.tf
```diff
@@ -0,0 +1,82 @@
...
+  bucket = var.legacy_claims_bucket_name
+}
+
+data "aws_dynamodb_table" "legacy_claims_table" {
+  name = var.legacy_claims_table_name
+}
+
+data "aws_dynamodb_table" "legacy_block_index_table" {
+  name = "${terraform.workspace == "prod" ? "prod" : "staging"}-${var.legacy_block_index_table_name}"
+}
```
alanshaw, line 29: If you want to configure your own it'll always be given the prefix "staging". How about this instead?:

```suggestion
variable "legacy_block_index_table_name" {
  description = "The name of the legacy block index DynamoDB table"
  type = string
  default = "${terraform.workspace == "prod" ? "prod" : "staging"}-ep-v1-blocks-cars-position"
}

data "aws_s3_bucket" "legacy_claims_bucket" {
  bucket = var.legacy_claims_bucket_name
}

data "aws_dynamodb_table" "legacy_claims_table" {
  name = var.legacy_claims_table_name
}

data "aws_dynamodb_table" "legacy_block_index_table" {
  name = var.legacy_block_index_table_name
}
```

deploy/app/legacyclaims.tf
```diff
@@ -0,0 +1,82 @@
...
+
+data "aws_iam_policy_document" "lambda_legacy_dynamodb_get_document" {
+  statement {
+    actions = [
+      "dynamodb:GetItem",
```
alanshaw, line 34:
```suggestion
      "dynamodb:Query",
```

alanshaw: I don't believe either of these tables are accessed using `GetItem`. You might want to rename `_get` in these definitions to `_query`?

alanshaw (changes requested):

deploy/app/legacyclaims.tf
volmedo: putting the conditional in the variable rather than the data source was the first thing I tried, but terraform complains about it.

In any case, using staging for anything not prod is intended. Since it's a resource not being set up within this configuration, it doesn't really make sense to make it dependent on the workspace used when deploying this service.

BTW, I'm failing to see how your suggestion behaves differently than the proposed change.

alanshaw (approved):

deploy/app/legacyclaims.tf
volmedo: ah, I see what you mean now. Variables only allow literal values for their `default` attribute. I implemented a workaround using `locals`.

PR #71, opened 2024-12-18
fix: API gateway domain name same as cert

volmedo:
When `terraform.environment` is `prod`, we are creating a cert for the API for `indexer.storacha.network`, but we then set the domain name to `prod.indexer.storacha.network`.

This PR sets the domain name to whatever name was used for the cert.


alanshaw (approved):
LGTM

PR #80, opened 2024-12-21
Refactor legacy claims service to use composable content claims finder

hannahhoward:
# Goals

- Solve caching issues with #57 
- Simplify caching logic -- only use one instance we know works

# For discussion

ContentClaimsFinder isn't perfect because we still have the url.Url third parameter -- though arguably this could be initialized with SimpleFinder and content claims URL for the LegacyClaims instead of the NotFoundFinder

Still this feels composable and clean and easy to transfer to the actual finder


alanshaw (approved):
👌

PR #82, opened 2025-01-08
feat(legacy): add carpark fallback

### LINKED ISSUE #63: Legacy claims should materialize directly from R2 when location claim not found
Equivalent of the code here https://github.com/storacha/blob-fetcher/blob/4566f96598e274dba18f075c6bbcf55fc8f7841b/src/locator/content-claims.js#L138-L152

hannahhoward:
# Goals

Return location claims that were never published properly on the blob protocol from ~May - November 2024
fix #63

# Implementation

Essentially this ports https://github.com/storacha/blob-fetcher/pull/14

- When no legacy claims are found in the legacy claims bucket, check if the cid in question has an uploaded blob in carpark (via head request)
- if present, generate a location claim on demand, in much the same way we do for data pre-blob protocol data (reading from the blob-index table)
- however, here we fold into the normal indexing query process, and the way we accomplish this is via identity CIDs. Which is... a bit funky, but seems the simplest mechanism for making everything work together :)

# For Discussion

TODO:
- test for bucketFallbackMapper
- maybe integration test the whole thing

### DIFF
--- a/pkg/aws/bucketfallbackmapp.go
+++ b/pkg/aws/bucketfallbackmapp.go
@@ -1,0 +1,82 @@
+package aws
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/ipfs/go-cid"
+	multihash "github.com/multiformats/go-multihash"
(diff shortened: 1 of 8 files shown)


alanshaw (approved):
🤯 wow, cool solution!

PR #83, opened 2025-01-10
chore(terraform): deploy staging to a different region

### LINKED ISSUE #54: staging `indexing-service` should point to staging `content-claims` table and bucket
Currently, `indexing-service` is configured to use the production `content-claims` table and bucket to fetch legacy claims. In staging deployments, it should point to the staging `content-claims` infra.

This might be tricky as right now both production and staging deployments of `indexing-service` are done to the same AWS region. `content-claims` deployments, however, use different regions for prod and staging. It might be needed (and, actually, recommended) to make `indexing-service` staging deployments to happen to the same region used for `content-claims` staging deployments (which, in turn, might require cleaning the current staging infra).

volmedo:
Resolves #54 

Most of our services get staging and production environments deployed to different regions. This PR enables this separation for the `indexing-service`.

The changes also make the staging `indexing-service` use the staging `content-claims` resources, instead of always pointing to the production ones as it does right now.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,56 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "*"
+    paths:
+      - ".github/workflows/deploy.yml"
+      - "cmd/**"
+      - "deploy/**"
+      - "pkg/**"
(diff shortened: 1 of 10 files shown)


alanshaw (approved):
🚀 thanks for doing this!

PR #85, opened 2025-01-11
Fix/claim fallback cid

hannahhoward:
# Goals

Fix issues with #82 

# Implementation

My identity CID approach has a weakness -- the actual CID used to look up claims is the link for the root block, not cid of the whole claim serialized to CAR.

This PR converts to using root links as the identity cid, rather than the whole serialized claim. Fortunately, this is roughly equivalent since the synthesized claim has only one block.

This means that a fallback using an identity CID will work all the way through the query process.

You can see an alternate approach in the commit history continues to use the identity CID of the serialzied CAR but modifies the legacy claims mapper so that it passes on this CID correctly. Ultimately I decided this approach was more "correct" and less messy, though it means some slightly odd reconstruction of delegations to maintain the identity CID.

query before

```bash
> go run ./cmd query -u https://hannah.indexer.storacha.network baguqeera2tvekmqau6baz36o5cudqiekmcb2boqspofdjip3b3ib2hn6k6qq
2025-01-10T17:03:48.214-0800	FATAL	cmd	cmd/main.go:179	querying service: http request failed, status: 500 Internal Server Error, message: processing query: failure response fetching claim. URL: https://prod-content-claims-bucket-claimsv1bucketefd46802-1mqz6d8o7xw8.s3.us-west-2.amazonaws.com/bafyreie4t46e4jhj5cxuhxedghgvackmz5qjfgbnu5jubgfpn5t3xh5f54/bafyreie4t46e4jhj5cxuhxedghgvackmz5qjfgbnu5jubgfpn5t3xh5f54.car, status: 403 Forbidden, message: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>AAQA544AX7932BM3</RequestId><HostId>CdB5O9RC4GzyU1UTWqCCX+t5m6u6XeoNaVKB/alTLLsSbPV1LEo21u03bEilxaWwx5EFmsp2FWQ=</HostId></Error>
```

query after

```bash
> go run ./cmd query -u https://hannah.indexer.storacha.network baguqeera2tvekmqau6baz36o5cudqiekmcb2boqspofdjip3b3ib2hn6k6qq

Query:
  Hashes (1):
    zQmcfnLAk4i1bZETqNde8qL6fjFA5CNikbNkdK6qQLdj67a

Results:
  Claims (2):
    bafyreibkvewseedfog6lbbkaijzbo2ggbz2nwna2frntxsxdvcceiwfcoy
      Type:
        assert/index
      Content:
        baguqeera2tvekmqau6baz36o5cudqiekmcb2boqspofdjip3b3ib2hn6k6qq
      Index:
        bagbaieradicy56yjedke4kaqw2aq34scek4wzoytw37suh34bfi5fdskubgq
    bafyreig4f3n73zzv5plhngchcejkanmvodjvery5ztyxwd4tjwzudbkpke
      Type:
        assert/location
      Content:
        zQmQ6EE3NqfJTaC5DtxUCLbtQ9bVoVGss8PAqP51MLdJ6uA
      Locations:
        https://carpark-prod-0.r2.w3s.link/zQmQ6EE3NqfJTaC5DtxUCLbtQ9bVoVGss8PAqP51MLdJ6uA/zQmQ6EE3NqfJTaC5DtxUCLbtQ9bVoVGss8PAqP51MLdJ6uA.blob
      Range: 0-379

  Indexes (1):
    bagbaiera23wlwcndsnbnp6g4f4a53obzwrjlygai4pmxqodfzgjyd4n5kanq
      Content:
        baguqeera2tvekmqau6baz36o5cudqiekmcb2boqspofdjip3b3ib2hn6k6qq
      Shards (1):
        zQmSQ26wcoCYcdMpwUbMkaZNfow3Y8B3iHjCjNkCqwkSAGT
          Slices (2):
            zQmSQ26wcoCYcdMpwUbMkaZNfow3Y8B3iHjCjNkCqwkSAGT @ 0-186
            zQmcfnLAk4i1bZETqNde8qL6fjFA5CNikbNkdK6qQLdj67a @ 98-186
```


hannahhoward:
upon further investigation, there's an issue with the final approach -- a "Data Integrity Error" when you can .Data() on the claim, because the blocks in UCAN are required to be CBOR encoded. I've reverted to my initial approach of just passing the claim CID

volmedo:
To confirm I understand this correctly:
1. claims generated on-demand are coming out of the `BucketFallbackMapper` with a CID built as an identity CID
1. however, we then synthetize `ProviderResult`s from them we use the claim's root link as the CID in the metadata, rather than the generated identity CID
1. when the client looks for the claim using the CID in the `ProviderResult`'s metadata, it gets nothing because we are identifying the claim by the identity CID, and not the root link CID
1. solution: put the identity CID in the metadata so that it points in the right direction

does that sound right?

And another question: the claims generated by the `BucketFallbackMapper` are ephemeral, in the sense that they will live in the cache until they expire, and then they'll need to be generated on-demand again, is that correct?

hannahhoward:
@volmedo yes! everything about your understanding is correct.

PR #86, opened 2025-01-12
refactor(providerindex): accept multiple legacy options

hannahhoward:
# Goals

Get a full roundtrip to work with freeway -- these are all the additional changes required

# Implementation

- update go-capabilities to include fixes from https://github.com/storacha/go-capabilities/pull/6 (which had not been updated in a minute, so there are a few changes required)
- modify the legacy claims finder to work with multiple claim mappers, and be sensitive to target claims. the reason for this is as follows -- I had assumed by fallback mapper could simply run on top of the base legacy mapper and run when the legacy mapper returns no claims. however, in practice, we end up in a situation where we are specifically looking for a Shard CID location claim, and the base legacy mapper returns a Shard CID EQUALS claim. So we need to have that filtered out and then run my fallback mapper anyway. This ends up making much more sense to do at the legacy claims finder level than in the fallback mapper.

# For Discussion

Still need to add some new test cases for the legacy claims finders now that it uses multiple mappers & handles filtering


alanshaw (approved):
Incredible debugging!

pkg/service/providerindex/legacy.go
```diff
@@ -55,16 +59,29 @@ func NewLegacyClaimsStore(contentToClaimsMapper ContentToClaimsMapper, claimStor
...
+func (ls LegacyClaimsStore) Find(ctx context.Context, contentHash multihash.Multihash, targetClaims []multicodec.Code) ([]model.ProviderResult, error) {
+	for _, mapper := range ls.contentToClaims {
+		claims, err := ls.findOne(ctx, contentHash, targetClaims, mapper)
+		if err != nil {
+			return nil, err
+		}
+		if len(claims) > 0 {
+			return claims, nil
+		}
+	}
```
volmedo, line 79: the results of this logic will depend on the order the mappers are provided in the slice. The results being returned will be the ones given by the first mapper returning any result.

Not saying it's wrong, but it may be worth stating it clearly somewhere (in the comment for `NewLegacyClaimsStore`, maybe?).

pkg/service/providerindex/legacy_test.go
volmedo, line 1: maybe add a test for correct filtering of claims according to `targetClaims`?

pkg/service/providerindex/legacy.go
```diff
@@ -55,16 +59,29 @@ func NewLegacyClaimsStore(contentToClaimsMapper ContentToClaimsMapper, claimStor
...
+	return nil, nil
```
volmedo, line 80: np: while equivalent in this case, returning an empty array feels more idiomatic IMO
```suggestion
	return []model.ProviderResult{}, nil
```

pkg/service/providerindex/legacy.go
```diff
@@ -55,16 +59,29 @@ func NewLegacyClaimsStore(contentToClaimsMapper ContentToClaimsMapper, claimStor
...
+}
+
+func (ls LegacyClaimsStore) findOne(ctx context.Context, contentHash multihash.Multihash, targetClaims []multicodec.Code, mapper ContentToClaimsMapper) ([]model.ProviderResult, error) {
```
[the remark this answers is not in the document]
hannahhoward: it's checking one mapper when the function above checks many

pkg/service/providerindex/legacy.go
hannahhoward: yes that's the goal, but it's absolutely worst stating

pkg/service/providerindex/legacy_test.go
hannahhoward: yea

pkg/service/providerindex/legacy.go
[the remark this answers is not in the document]
volmedo: got it. The very first thing that came to my mind was "find one result". I guess I'm too used to datastore APIs :)

volmedo:
I tidied some things up, updated the branch with the latest main and added tests to confirm handling of different mappers and claim filtering work as expected.

This is RFAL.

alanshaw (approved):
Still LGTM

PR #87, opened 2025-01-14
feat(query): add tests for IndexingService.Query function

volmedo:
Resolves https://github.com/storacha/project-tracking/issues/171

Adds tests for the main `IndexingService.Query` function. The happy path test is useful to understand the behavior of the service fetching the different types of claims.

**Note for reviewers:** the PR is not as big as it seems. I changed some interfaces dealing with `url.URL` to use `*url.URL`, which makes client code a bit more readable and is more idiomatic. A lot of files were touched just to update the type of the parameter/variable.

## Considerations

I think the happy path test is the most valuable. I also added some basic tests for error conditions in the main dependencies. I thought of adding a test confirming the function filters non-relevant claims, but that is implemented by the `providerindex` package and tested there.

Please let me know if you think I'm missing any other interesting tests.

## Known issues

The addition of `MockProviderIndex` causes an import cycle in the tests of the `providerindex` package. These tests import the `mocks` package and the new mock imports the `providerindex` package to implement the mock.

I think the best solution would be to change the mock generation configuration so that mocks are created in the same package as the interfaces they are mocking. That would avoid the cycle while allowing tests to live in the same package as the functions they are testing (which in turn enables testing non-exported members).

I'll take care of that in a separate PR (#88).


pkg/internal/jobwalker/singlewalk/singlewalk.go
```diff
@@ -41,12 +41,15 @@ func SingleWalker[Job, State any](ctx context.Context, initial []Job, initialSta
 			return state.Access(), ctx.Err()
 		default:
 		}
-		next := initial[len(initial)-1]
-		initial = initial[:len(initial)-1]
-		handler(ctx, next, func(j Job) error {
+		next := stack[len(stack)-1]
+		stack = stack[:len(stack)-1]
+		err := handler(ctx, next, func(j Job) error {
```
alanshaw, line 46: 👌 good catch

PR #91, opened 2025-01-17
feat: add diagram of resources deployed in AWS

volmedo:
Add a quick diagram of what a deployment of the service looks like in AWS.


fforbeck (approved):
Very useful! Let's add it to the repo :)

PR #96, opened 2025-01-24
feat: add support for OpenTelemetry

volmedo:
Ref: https://github.com/storacha/project-tracking/issues/246

This PR adds support for OpenTelemetry instrumentation.

The traces are quite basic for now, with a single span for the lambda handler that is invoked, which gives little info. I'm following up with another PR that adds more details about outgoing requests to other services.

## Implementation details

The main issue I faced with this was related with lambda lifecycles. It seems that a lambda can be frozen at any given time once the handler has been executed. OpenTelemetry defaults to using batching and asynchronous processing of traces and spans, which doesn't play well with the lambda being sent to sleep potentially at any moment.

It is indeed possible to export traces directly from the handler logic, but it needs to be done synchronously. As this would have an impact on request processing times, the recommended practice for production deployments is to offload trace batching and exporting to an independent collector process.

This is what the [AWS Distro for OpenTelemetry Lambda layer](https://aws-otel.github.io/docs/getting-started/lambda) offers. It is a lite version of the OpenTelemetry collector that exports telemetry data to X-Ray by default, but can be configured to export to other backends.

Configuration is a bit cumbersome because a configuration file is required, that's why I added that configuration file and changed the Makefile and the terraform script so that it gets bundled with the lambda binary.


cmd/lambda/getclaim/main.go
```diff
@@ -2,20 +2,48 @@ package main
...
+	"go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda"
 )
 
 func main() {
-	config := aws.FromEnv(context.Background())
-	service, err := aws.Construct(config)
+	cfg := aws.FromEnv(context.Background())
+
+	// an empty API key disables instrumentation
+	if cfg.HoneycombAPIKey != "" {
```
hannahhoward, line 19: noticing this code is essentially shared between every command. you could probably define a withTelemetry command that took a makeHandler and a cfg and then just ran this conditional.

hannahhoward (approved):
on suggestion, but non-blocking. LGTM

cmd/lambda/getclaim/main.go
volmedo: good point! I'll refactor it in my next PR.

PR #97, opened 2025-01-27
chore: upgrade go-capabilities lib to latest version

volmedo:
Upgrade [storacha/go-capabilities ](https://github.com/storacha/go-capabilities) to the latest version that updates the schema of EqualsCaveats. This should allow the service to correctly deserialize equals claims.

In the latest version of go-capabilities some things changed places so some imports needed to be adjusted.


alanshaw (approved):
I don't think there's anything wrong here but I think this is less readable than it was before. While `cassert` is consistent with `ctypes` I think that the default name (`assert`) would be more readable.

Perhaps also re-exporting `FromHash` and `Digest` from `pkg/assert` in `go-capabilities` would reduce the amount of change here - it is _very_ likely if you're using the assert types that you're going to need that function and type.

PR #100, opened 2025-01-30
feat: allow specifying query type in HTTP API

### LINKED ISSUE #94: Allow specifying query type on service interface and HTTP
# What

As a client of the indexer making HTTP calls, I should be able to specify query type ("standard", "index_or_location", "location") as an HTTP parameter to a query request

Note we are exposing the three query types used by the query execution logic, not an arbitrary filter for claim types

# Why

As a client, there are times when I have partial information about a CID/Multihash and I want to start a query "in the middle" -- jumping over earlier parts of the request. Most typically, this could mean I have an index for a cid and I know what shard its in, but I don't yet have a location claim for the shard.

# Implementation

- Make query type a public type
- Add it to the Query struct
- pass it to the query logic in the service handler
- add a query parameter to the http handler
- parse and validate the query parameter
- pass it to the Query service
- add support in the http client

# Acceptance Criteria

For a multihash with a equals claims, an index claim and a location claim:

`GET /claims?multihash=~multihash~` should return all three claims ('standard' is default)

`GET /claims?multihash=~multihash~&kind=index_or_location` should return only the index and location claim

`GET /claims?multihash=~multihash~&kind=location` should return only the location claim

volmedo:
💯 

closes #94

This PR adds a new `kind` query parameter to the `/claims` endpoint that indicates which claim types should be returned. It has three possible values:
- `standard`: this is the default, all three types of claims (location, index and equals) are returned
- `location`: only location claims are returned
- `index_or_location`: location and index claims are returned

Aside from adding support in both the server and the client, a new `type`/`t` parameter is added to the `indexer` CLI so that different query types can be performed.


alanshaw (changes requested):
Is it "type" or "kind"? Please can we be consistent.

Why would we not add this to the `Match` parameter of the query? It's essentially a filter like space DID, no?

volmedo:
agree on the "type" vs. "kind" thing. I'll set everything to "type" as it's a simpler change and nothing consumes the new parameter yet.

about adding the type to `Match`, I think the current implementation aligns better with this requirement in #94:
> Note we are exposing the three query types used by the query execution logic, not an arbitrary filter for claim types

I also think "match" has the meaning of "values to match your query against", so I'd put claim types rather than a query type. But then the API would be too flexible. I don't have a strong opinion, though, so let me know if you prefer this to be added to `Match`.

alanshaw (approved):
LGTM

PR #101, opened 2025-01-31
feat: enable point in time recovery and deletion protection in prod

alanshaw:
Enables deletion protection for production DynamoDB tables so we don't accidentally delete them. Also enables point in time recovery...it might be possible to recreate these tables from the chain data but it will probably be better if we can just restore them.


volmedo (approved):
I'm taking a look at the test failure. It is indeed flaky.

PR #102, opened 2025-01-31
flaky test

volmedo:


volmedo:
the traces show the issue was a race condition starting the server and running the test (see [this failing run of the test](https://github.com/storacha/indexing-service/actions/runs/13074250953/job/36492928500?pr=102#step:15:72)).

the solution is implemented in #104

PR #104, opened 2025-02-01
fix(tests): flaky test in BucketFallbackMapper tests

### LINKED ISSUE #103: `TestBucketFallbackMapper/200_status_code_on_head_generates_claim` is flaky
The test fails sometimes, only on Ubuntu.

Example failing run: https://github.com/storacha/indexing-service/actions/runs/13030877549/job/36349708913#step:15:459

volmedo:
Fixes #103 

The original implementation launched an HTTP server for the test concurrently in a different goroutine. This poses a race condition between starting the server and running the tests. Sometimes, the test was run before the server was ready and the test failed.

Finding the cause has been tricky because, in case of an error in the HEAD request, `BucketFallbackMapper` swallows the original error and returns `ErrKeyNotFound`, masking the real cause of the issue. This error handling has been improved to avoid this.

To fix the test, the regular HTTP server being used has been replaced with an `httptest.Server`. It simplifies the test code as handling its shutdown is easier, but the reason it fixes the issue is that `httptest.Server` prepares the network listener synchronously, and then starts the server concurrently. That means that, even in the case the test is run before the server is ready to server requests, the listener will be able to buffer them and prevent the test from failing.

### DIFF
--- a/pkg/aws/bucketfallbackmapper.go
+++ b/pkg/aws/bucketfallbackmapper.go
@@ -48,9 +48,13 @@ func (cfm BucketFallbackMapper) GetClaims(ctx context.Context, contentHash multi
 	}
 
 	resp, err := cfm.httpClient.Head(cfm.bucketURL.JoinPath(toBlobKey(contentHash)).String())
-	if err != nil || resp.StatusCode < 200 || resp.StatusCode >= 300 {
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


volmedo:
yeah, I guess each test case having its own server aligns with the guideline of isolating the tests as much as possible, but the handler is always the same, so I think it's ok in this case.

PR #108, opened 2025-02-05
fix: region for legacy claims bucket

alanshaw:
If your AWS deployment region is set to `us-west-2` for example, then your deployment will fail because the staging claims bucket is actually in `us-east-2`.

### DIFF
--- a/deploy/app/legacyclaims.tf
+++ b/deploy/app/legacyclaims.tf
@@ -36,14 +36,14 @@ locals {
     inferred_legacy_block_index_table_name = var.legacy_block_index_table_name != "" ? var.legacy_block_index_table_name : "${terraform.workspace == "prod" ? "prod" : "staging"}-ep-v1-blocks-cars-position"
 }
 
-data "aws_s3_bucket" "legacy_claims_bucket" {
-  provider = aws.legacy_claims
-  bucket = local.inferred_legacy_claims_bucket_name
-}
-
 provider "aws" {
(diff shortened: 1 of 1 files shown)


deploy/app/legacyclaims.tf
```diff
@@ -37,6 +37,7 @@ locals {
 }
 
 data "aws_s3_bucket" "legacy_claims_bucket" {
+  provider = aws.legacy_claims
   bucket = local.inferred_legacy_claims_bucket_name
 }
```
volmedo, line 42: np: maybe move the `provider` block below to before this `data` block, now that it is used here?

volmedo (commented):
that's weird, I thought S3 buckets had a looser relation with regions. I know they are associated with one, but they can be accessed from others as well, which doesn't happen with other resources.

anyway, this shouldn't happen because the idea is to only access the staging bucket from the staging region. So maybe it's actually a good thing that it fails? 😅. You should destroy your env in the prod region and recreate in the staging one.

alanshaw:
> So maybe it's actually a good thing that it fails? 😅.

Ha, yes but unintuative error. If we want to enforce deployment region to be the same as legacy claims region (which I think is reasonable) then we should explicitly codify that.

> You should destroy your env in the prod region and recreate in the staging one.

I didn't get that far but yes I was going to create in staging region.

volmedo (approved):

PR #110, opened 2025-02-05
feat: configurable principal mapping

alanshaw:
Echos the [principal mapping config in storage node](https://github.com/storacha/storage/blob/c234bbbc56915ec40ad3738607d008848f71c136/deploy/app/variables.tf#L89-L93).

Alllows services with did:web keys other than the hard coded defaults to use the indexer.

### DIFF
--- a/.env.local
+++ b/.env.local
@@ -5,3 +5,6 @@ TF_VAR_did=
 
 # optional - if you want to enable instrumentation (only Honeycomb is supported for now)
 # TF_VAR_honeycomb_api_key=
+
+# optional - JSON encoded mapping of did:web to did:key
+# TF_VAR_principal_mapping=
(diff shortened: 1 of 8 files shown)


volmedo (approved):
LGTM

It may make sense to add some brief information about the new configuration to the readme (an example could also be useful), but we can do that later.

alanshaw:
Agree, I'd like to propose that we actually move that information into the template `.env` file so we don't repeat ourselves or get out of sync with the vars that are available and the vars that are documented in the README. This is the direction I took in the storage node: https://github.com/storacha/storage/blob/feat/aws/deploy/.env.tpl

WDYT?

volmedo:
yeah, that works too

PR #115, opened 2025-02-07
fix: blob index compatibility

alanshaw:
JS blob index (the ones we create on the client) encodes the root block as the first block in the CAR not the last.

This PR restores parity with the JS blob index, allowing indexes created using Go to serialize to the same bytes as indexes created by JS and by virtue hash to the same CID:

BEFORE:

Screenshot 2025-02-07 at 15 28 12

AFTER:

Screenshot 2025-02-07 at 15 27 49


hannahhoward (approved):
LGTM

PR #116, opened 2025-02-10
feat(elasticache): open one port, switch to valkey

hannahhoward:
# Goals

Noticing high connect times for redis in traces -- probably just first query but still https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/wwe-troubleshooting.html recommends having both ports open in case a client tries to connect to both

Also, we need to switch to valkey before we turn it on cause it's almost 1/3 cheaper per unit.

# Implementation

- open both ports
- upgrade AWS provider to get valkey support
- move to valkey
- set a dummy password on the disabled default user -- no-password-required apparently not supported for valkey
- fix some downstream effects of create_before_destroy being applied at top of stack (shouldn't be as create_before_destroy will not work for many resources)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,11 @@ deploy/app/.terraform:
 
 init: deploy/app/.terraform .tfworkspace
 
+.PHONY: upgrade
+
+upgrade:
+	tofu -chdir=deploy/app init -upgrade
+
 .PHONY: validate
 
 validate: deploy/app/.terraform .tfworkspace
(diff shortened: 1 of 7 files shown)


alanshaw (approved):

volmedo:
tested this in my staging env and everything looks good, merging.

PS: I see a similar pattern of the first query to the cache being slow, it looks like it's due to warm up indeed.

PR #117, opened 2025-02-10
fix: allow skipping publish of existing advert

alanshaw:
If one or multiple users upload the same blob multiple times the `assert/index` invocation will create an advert that already exists. This will cause `ipni-publisher` to return `ErrAlreadyAdvertised` and the invocation fails. This PR ignores `ErrAlreadyAdvertised` because, that's fine, whatever.

It also adds a mutex around publish per https://github.com/storacha/ipni-publisher/issues/3


volmedo (commented):
left a couple nitpicks, but looks good

PR #119, opened 2025-02-11
feat: increase concurrency

volmedo:
Traces in prod show queries taking around 9 secs to complete. It is clear from taking a look at the spans that the concurrency level is being hit and an increase could improve query time.

I'm bumping it to 15 for now to test how it goes. We might want to choose an even higher number. Concurrency is based in goroutines, which are very efficient, so it should be possible to go much higher. The limiting factor would probably be networking.


alanshaw (approved):

hannahhoward:
FYI this may not make a huge difference cause the actual query logic is mostly sequential, unless you are looking for multiple hashes at once.

PR #121, opened 2025-02-12
feat(provider-index): parallelize claim lookups

### LINKED ISSUE #120: Parallelize ProviderIndexService Find Method
## What
At present, `ProviderIndexService` looks up claims sequentially:
1. Check cache -> return if found, otherwise...
2. Query IPNI -> return if found, otherwise...
3. Query the Legacy Claims Index -> return if found, otherwise not found.

The problem? Step 2 (IPNI) has to finish before Step 3 starts. If IPNI doesn’t have the claim but Legacy does, we waste time waiting on IPNI.

## Proposal
Run IPNI and Legacy Claims Index in parallel to speed things up. The key rule:
- Prioritize IPNI -> If IPNI has results, use them.
- Only use Legacy if IPNI is empty -> No mixing results.
- Still cache everything to keep future lookups fast with priority to IPNI results.

This keeps IPNI as the primary source but avoids unnecessary delays when it's empty.

## Why?
Right now, getProviderResults forces a strict step-by-step lookup, which slows things down. Running both queries in parallel cuts down on wait time while keeping the existing behavior intact.

This change makes lookups faster, keeps caching as-is, and doesn’t mess with the logic—just makes it more efficient

frrist:
via 656fa6769c5dad56e2a99590930f09dd2f324e1b
- ProviderIndexService.getProviderResults now queries IPNI and legacy claims concurrently using channels. If IPNI returns a valid result, the legacy lookup is canceled.
- errors from both queries are joined to provide better context when neither returns data.
- tests have been updated to cover these new scenarios.
- closes #120 

via ae21e4484301fc1f3f8f869f256e52d9e27ae560 
- adds a few more events to the span for the operation be modified allowing timing of various operations to be inferred. If we prefer this change be landed separately (or not at all!) happy to revert it.

### DIFF
--- a/pkg/service/providerindex/providerindex.go
+++ b/pkg/service/providerindex/providerindex.go
@@ -22,6 +22,7 @@ import (
 	"github.com/storacha/indexing-service/pkg/telemetry"
 	"github.com/storacha/indexing-service/pkg/types"
 	"github.com/storacha/ipni-publisher/pkg/publisher"
+	"go.opentelemetry.io/otel/trace"
 )
 
 var log = logging.Logger("providerindex")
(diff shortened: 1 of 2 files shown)


hannahhoward (approved):
LGTM

frrist:
@volmedo great idea :smile: 
- I've added a couple new tests via 13931b0d74bf9f12248407d4c4ee07e2216cdee9 that cover the case when IPNI/Legacy return a result and Legacy/IPNI return an empty result (but no error). 
- Coverage for neither returning a result nor error is (thankfully) already well covered here: https://github.com/storacha/indexing-service/blob/9b1c5d6edae4292b132fbd1f7d37b1ec5eafce2b/pkg/service/providerindex/providerindex_test.go#L127-L145

volmedo:
thanks for adding the extra tests @frrist! 💯 

A nitpick that we may want to implement at some point: I spotted a couple comments that I think should read different (in lines [321](https://github.com/storacha/indexing-service/blob/61a312d0c97dfe6b0a6ab9bf2bc46c010b4671cd/pkg/service/providerindex/providerindex_test.go#L321) and [371](https://github.com/storacha/indexing-service/blob/61a312d0c97dfe6b0a6ab9bf2bc46c010b4671cd/pkg/service/providerindex/providerindex_test.go#L371)). They mention that an error is returned when it's in fact an empty result.

PR #126, opened 2025-02-13
Override data bucket url

hannahhoward:
# Goals

Enable setting the carpark to use for the LEGACY_DATA_BUCKET_URL. This simply completes the set of vars needed to fully complete custom configuration of legacy lookups

Useful for local testing downstream services like Freeway against prod data, but without using the prod service

# Implementation

- add a var for legacy data bucket url
- use legacy data bucket url when configured

### DIFF
--- a/.env.local
+++ b/.env.local
@@ -8,3 +8,10 @@ TF_VAR_did=
 
 # optional - JSON encoded mapping of did:web to did:key
 # TF_VAR_principal_mapping=
+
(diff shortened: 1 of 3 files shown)


volmedo (commented):
LGTM but it looks like the first commit in the PR refers to changes that were already merged. I'm confused by the fact they appear in the diff because I can see them already in main 🤷🏻. Am I missing something?

hannahhoward:
@volmedo sorry forgot to rebase

PR #129, opened 2025-02-13
feat(redis): enable tuning of caching parameters

hannahhoward:
# Goals

Currently, our cache hit rate for some of our redis instances is really low, because it turns out 1 hour really isn't long enough to hold some of this data. There's not reason not to hold it longer -- we've allocated almost 10GB to these redis instances and we're currently using all of a couple MB.

Screenshot 2025-02-13 at 1 30 19 PM

# Implementation

- Enable passing a specific expiration parameter to each redis store
- Pipe all the way out to an AWS env var
- Up expirations times to:
  - 1 month on prod, 1 week on staging for providers (very small data)
  - 1 week on prod, 1 day on staging for claims
  - 1 day on prod, 1 hour on staging for indexes (larger data)

### DIFF
--- a/deploy/app/lambda.tf
+++ b/deploy/app/lambda.tf
@@ -70,10 +70,13 @@ resource "aws_lambda_function" "lambda" {
     variables = {
       	PROVIDERS_REDIS_URL = aws_elasticache_serverless_cache.cache["providers"].endpoint[0].address
         PROVIDERS_REDIS_CACHE = aws_elasticache_serverless_cache.cache["providers"].name
+        PROVIDERS_CACHE_EXPIRATION_SECONDS = "${terraform.workspace == "prod" ? 30 * 24 * 60 * 60 : 24 * 60 * 60}"
         INDEXES_REDIS_URL = aws_elasticache_serverless_cache.cache["indexes"].endpoint[0].address
         INDEXES_REDIS_CACHE = aws_elasticache_serverless_cache.cache["indexes"].name
+        INDEXES_CACHE_EXPIRATION_SECONDS = "${terraform.workspace == "prod" ? 24 * 60 * 60 : 60 * 60}"
         CLAIMS_REDIS_URL = aws_elasticache_serverless_cache.cache["claims"].endpoint[0].address
         CLAIMS_REDIS_CACHE = aws_elasticache_serverless_cache.cache["claims"].name
+        CLAIMS_CACHE_EXPIRATION_SECONDS = "${terraform.workspace == "prod" ? 7 * 24 * 60 * 60 : 24 * 60 * 60}"
         REDIS_USER_ID = aws_elasticache_user.cache_iam_user.user_id
         IPNI_ENDPOINT = "https://cid.contact"
         PROVIDER_CACHING_QUEUE_URL = aws_sqs_queue.caching.id
--- a/pkg/aws/service.go
+++ b/pkg/aws/service.go
@@ -9,6 +9,7 @@ import (
 	"maps"
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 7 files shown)


pkg/aws/service.go
```diff
@@ -45,28 +47,40 @@ func mustGetEnv(envVar string) string {
 	return value
 }
 
+func mustGetInt(envVar string) int64 {
+	stringValue := mustGetEnv(envVar)
+	value, err := strconv.ParseInt(stringValue, 10, 64)
+	if err != nil {
+		panic(fmt.Errorf("parsing env var %s to int: %w", envVar, err))
```
frrist, line 54: suggestion: we may want to include the `stringValue` returned that was expected to be an int, although I suspect this is a _very_ rare panic.

deploy/app/lambda.tf
```diff
@@ -70,10 +70,13 @@ resource "aws_lambda_function" "lambda" {
     variables = {
       	PROVIDERS_REDIS_URL = aws_elasticache_serverless_cache.cache["providers"].endpoint[0].address
         PROVIDERS_REDIS_CACHE = aws_elasticache_serverless_cache.cache["providers"].name
+        PROVIDERS_CACHE_EXPIRATION_SECONDS = "${terraform.workspace == "prod" ? 30 * 24 * 60 * 60 : 24 * 60 * 60}"
         INDEXES_REDIS_URL = aws_elasticache_serverless_cache.cache["indexes"].endpoint[0].address
         INDEXES_REDIS_CACHE = aws_elasticache_serverless_cache.cache["indexes"].name
+        INDEXES_CACHE_EXPIRATION_SECONDS = "${terraform.workspace == "prod" ? 24 * 60 * 60 : 60 * 60}"
         CLAIMS_REDIS_URL = aws_elasticache_serverless_cache.cache["claims"].endpoint[0].address
         CLAIMS_REDIS_CACHE = aws_elasticache_serverless_cache.cache["claims"].name
+        CLAIMS_CACHE_EXPIRATION_SECONDS = "${terraform.workspace == "prod" ? 7 * 24 * 60 * 60 : 24 * 60 * 60}"
```
frrist, line 79: :heavy_check_mark: math is math-ing

pkg/aws/service.go
```diff
@@ -159,33 +173,36 @@ func FromEnv(ctx context.Context) Config {
...
-		ClaimStorePrefix:            os.Getenv("CLAIM_STORE_KEY_REFIX"),
-		LegacyClaimsTableName:       mustGetEnv("LEGACY_CLAIMS_TABLE_NAME"),
-		LegacyClaimsTableRegion:     mustGetEnv("LEGACY_CLAIMS_TABLE_REGION"),
-		LegacyClaimsBucket:          mustGetEnv("LEGACY_CLAIMS_BUCKET_NAME"),
-		LegacyBlockIndexTableName:   mustGetEnv("LEGACY_BLOCK_INDEX_TABLE_NAME"),
-		LegacyBlockIndexTableRegion: mustGetEnv("LEGACY_BLOCK_INDEX_TABLE_REGION"),
-		LegacyDataBucketURL:         mustGetEnv("LEGACY_DATA_BUCKET_URL"),
-		HoneycombAPIKey:             os.Getenv("HONEYCOMB_API_KEY"),
-		PrincipalMapping:            principalMapping,
+		ProvidersCacheExpirationSeconds: mustGetInt("PROVIDER_CACHE_EXPIRATION_SECONDS"),
```
frrist, line 176: blocking: PROVIDER**S**_CACHE_EXPIRATION_SECONDS

frrist (approved):
One blocking comment wrt env var misspelling, LGTM otherwise.

frrist:
Question: What's our motivation for using Time-Based Expiry rather than Capacity-Based Expiry (LRU or LFU) for this cache?

hannahhoward:
Capacity based expiry is already built into AWS. :) (and will expire those with expiries first -- goodness I hope I configured that right though -- should probably go back and review :P)

hannahhoward:
ah yes -- https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/Scaling.html#:~:text=ElastiCache%20Serverless%20also%20supports%20a,from%20Replica%20using%20READONLY%20connections).

"When you set a cache data storage maximum and your cache data storage hits the maximum, ElastiCache will begin evicting data in your cache that has a Time-To-Live (TTL) set, using the LRU logic."

The reason for setting a TTL at all is that we want to distinguish between stuff that's already on IPNI and stuff we actually pre-cache on the indexer while it gets to IPNI. we want the pre-cache stuff to have no expiration until we get to IPNI, so it doesn't get deleted.

Also our capacity is 10GB but we get charged by the GB for what we actually use, so if our hit ration is high even with expiry, it's better to not use all 10GB at most times

PR #138, opened 2025-02-21
fix: move block index table legacy service to regular legacy claims flow

### LINKED ISSUE #135: Move Block Index Table Legacy Provider into the normal legacy provider flow using identity CIDs
# What

Currently we have a few different ways to find legacy claims:
- look in the content claims DB
- HEAD request to R2 for blob CIDs
- look at the blob index table

Of these the first two are offered as legacy options in the providers index, while the last one stands outside the entire indexing query process as an alternate flow.

The challenge of being part of the "alternate flow" is that nothing is cached in redis. Moreover, since this flows results are only used AFTER we know the normal flow did not produce results, a typical response time for this case looks like this:

Image

What happened here? Well, we ran the regular flow and the alternate flow in parallel, but we only used the alternate flow AFTER we got results from the regular flow. The regular flow took about 200ms to query IPNI, query the content claims, all head request R2 -- all to find nothing. Meanwhile the block index table had the results. Since the block index table is the ONLY resource we have for store protocol, this describes about 80% of calls. The real kicker is that nothing is cached, so the same call takes just as long the second time.

# Solution

Instead of an entire alternate flow, put the block index table lookup in as a 3rd legacy claims lookup after content claims and R2 head checks.

But how? Likely we did this approach originally because at the time, we didn't have a solution for the fact that these claims are synthetic -- i.e. they do not exist in any database, which made creating the flow from provider -> claim calls difficult. However, since then we came up with a solution which we used for the R2 Head requests -- putting the full content claims in an identity CID.

By getting rid of the alternate flow, and simply making the block index table a third legacy claims lookup that functions similar to https://github.com/storacha/indexing-service/blob/main/pkg/aws/bucketfallbackmapper.go , we can get these queries into the caching flow and get our cache hit rate up.

volmedo:
Resolves #135 

This PR takes the logic from the block index table legacy service and turns it into a ContentToClaimsMapper. Then adds that mapper to the list of mappers used by the regular legacy claims flow so that its results are handled and cached the same way as with the results from other mappers.


alanshaw (changes requested):

pkg/aws/service.go
```diff
@@ -243,7 +249,7 @@ func Construct(cfg Config) (types.Service, error) {
 		construct.WithPublisherStore(publisherStore),
 		construct.WithStartIPNIServer(false),
 		construct.WithClaimsStore(claimBucketStore),
-		construct.WithLegacyClaims([]providerindex.ContentToClaimsMapper{legacyClaimsMapper, bucketFallbackMapper}, legacyClaimsBucket, legacyClaimsURL),
+		construct.WithLegacyClaims([]providerindex.ContentToClaimsMapper{blockIndexTableMapper, legacyClaimsMapper, bucketFallbackMapper}, legacyClaimsBucket, legacyClaimsURL),
```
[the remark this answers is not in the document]
volmedo: good catch, I actually checked the original code to keep this behavior and then used the wrong order 🤦🏻

pkg/aws/blockindextablemapper.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	claimCids := make([]cid.Cid, 0, len(locs))
+	for _, loc := range locs {
+		claim, err := cassert.Location.Delegate(
+			bit.id,
+			bit.id,
+			bit.id.DID().String(),
+			loc,
+			delegation.WithExpiration(int(time.Now().Add(time.Hour).Unix())),
```
alanshaw, line 110: I believe the redis cache will hold this for 1 hour also...we should probably make this slightly more than an hour so we don't return a cached but expired delegation later...

alanshaw (approved):
LGTM other than the small tweak.

pkg/aws/blockindextablemapper.go
alanshaw: Ideally we'd link these in code so a change to the cache expiration does not break this in the future.

volmedo: interesting. I linked synthetized claim expiration to the expiration in the cache. I added a buffer of 10 minutes, do you think that's a proper amount?

hannahhoward: um... not an  hour any more. try a month, less of a problem :P

pkg/aws/service.go
```diff
@@ -233,6 +232,17 @@ func Construct(cfg Config) (types.Service, error) {
 			return []delegation.Option{delegation.WithExpiration(int(time.Now().Add(time.Hour).Unix()))}
 		},
 	)
+	blockIndexCfg := cfg.Config.Copy()
+	blockIndexCfg.Region = cfg.LegacyBlockIndexTableRegion
+	legacyBlockIndexStore := NewDynamoProviderBlockIndexTable(dynamodb.NewFromConfig(blockIndexCfg), cfg.LegacyBlockIndexTableName)
+
+	// allow claims synthethized from the block index table to live a bit longer after they are expired in the cache
+	// so that the service doesn't return cached but expired delegations
+	synthetizedClaimExp := redis.DefaultExpire + 10*time.Minute
```
hannahhoward, line 241: Blocking: the expiration time on the claims cache is configurable and is configured much greater in prod (1 week) than redis.DefaultExpire (1 hour). Please update (otherwise claims will expire before they expire from cache)

```suggestion
	synthetizedClaimExp := time.Duration(cfg.ClaimsCacheExpirationSeconds)*time.Second + 10*time.Minute
```

hannahhoward (changes requested):
LGTM except one important blocking change.

pkg/aws/blockindextablemapper.go
hannahhoward: nm I now see  the design and have suggested what I think is the right value in the setup code.

hannahhoward: it's actually a week in prod now.

pkg/aws/service.go
volmedo: didn't realize there was a setting for that, thanks!

I completely made up the 10 min buffer, does it make sense? I guess the claim will be used immediately by the gateway and 10 minutes is more than enough for it to complete the operation, right? Are there any other use cases that may need the claim to be valid for longer?

volmedo:
I'm bypassing the rules to merge because the requested changes are done

PR #139, opened 2025-02-24
fix: distributed tracing not working

### LINKED ISSUE #134: Distributed Tracing Not working
# What

Currently, trace context from freeway are not being properly read by indexer query lambda. Traces in honeycomb stop at theme moment it goes to the indexer from freeway, and show up as seperate traces in the indexer service

Likely this is due to some http parameter not being read by the API gateway and passed to the lambda, plus possibly not configuring the aws lambda otel library correctly. (perhaps need to configure an "EventCarrier" properly? or maybe we should be using XrayConfig tracer)

Docs:
https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda

Also, we should not trace unless there is a parent context -- currently we only run traces on Freeway every 1000 requests, and we should skip tracing when freeway itself is tracing -- otherwise we'll run out of trace space on honeycomb.

volmedo:
fixes #134 

Context propagation was not properly configured and incoming trace context was not being extracted from requests coming from the gateway.

Interestingly, telemetry was already configured to sample only requests that were sampled in the upstream service, but since context propagation was not working the result was that all requests were being sampled. Fixing context propagation also means that the indexer will only sample requests that are sampled in the gateway.

### DIFF
--- a/pkg/telemetry/otel.go
+++ b/pkg/telemetry/otel.go
@@ -89,29 +89,15 @@ func InstrumentLambdaHandler(handlerFunc interface{}) interface{} {
 // for the "headers" field in the event, which is populated by the AWS API Gateway. Therefore, this function will make
 // distributed tracing work only with HTTP-based lambdas that are triggered by the API Gateway.
 func jsonEventHeadersToCarrier(eventJSON []byte) propagation.TextMapCarrier {
-	ev := map[string]any{}
-	err := json.Unmarshal(eventJSON, &ev)
-	if err != nil {
+	var apiGatewayEvent struct {
+		Headers map[string]string `json:"headers"`
+	}
+
+	if err := json.Unmarshal(eventJSON, &apiGatewayEvent); err != nil {
 		return propagation.MapCarrier{}
 	}
 
-	// confirm the event has a headers field
-	if _, ok := ev["headers"]; !ok {
-		return propagation.MapCarrier{}
-	}
-
-	// confirm the headers field is a map
-	headers, ok := ev["headers"].(map[string]any)
-	if !ok {
-		return propagation.MapCarrier{}
-	}
-
-	headerMap := make(map[string]string, len(headers))
-	for k, v := range headers {
-		headerMap[k] = v.(string)
-	}
-
-	return propagation.MapCarrier(headerMap)
+	return propagation.MapCarrier(apiGatewayEvent.Headers)
 }
 
 func InstrumentHTTPClient(client *http.Client) *http.Client {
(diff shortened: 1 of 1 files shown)


pkg/telemetry/otel.go
```diff
@@ -65,11 +79,41 @@ func InstrumentLambdaHandler(handlerFunc interface{}) interface{} {
 
 	return otellambda.InstrumentHandler(
 		handlerFunc,
+		otellambda.WithEventToCarrier(jsonEventHeadersToCarrier),
```
frrist, line 82: Interesting: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda#EventToCarrier
> An EventToCarrier function defines how the instrumentation should prepare a TextMapCarrier for the configured propagator to read from. This extra step is necessary because Lambda does not have HTTP headers to read from and instead stores the headers it was invoked with (including TraceID, etc.) as part of the invocation event. If using the AWS XRay tracing then the trace information is instead stored in the Lambda environment.

So this is technically the "fix" for the issue, great find!

pkg/telemetry/otel.go
```diff
@@ -65,11 +79,41 @@ func InstrumentLambdaHandler(handlerFunc interface{}) interface{} {
...
+	if !ok {
+		return propagation.MapCarrier{}
+	}
+
+	headerMap := make(map[string]string, len(headers))
+	for k, v := range headers {
+		headerMap[k] = v.(string)
+	}
+
+	return propagation.MapCarrier(headerMap)
```
frrist, line 114: nit:

We are conforming to this event schema, yeah? https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html - if so, we could probably condense this a bit more to something like:
```go
    var event struct {
        Headers map[string]string `json:"headers"`
    }

    if err := json.Unmarshal(eventJSON, &event); err != nil {
        return propagation.MapCarrier{}
    }

    return propagation.MapCarrier(event.Headers)
```

frrist (approved):
LGTM

pkg/telemetry/otel.go
volmedo: yeah. It all comes from the fact that lambda handlers can have different shapes and receive many event schemas. Honestly, I would've expected `otellambda.InstrumentHandler` to take care of most of that automatically, at least for the most common events, such as the ones produced by the API Gateway.

Surprisingly enough, the headers are correctly transferred to the request injected into the lambda handler for HTTP requests, so using the standard OpenTelemetry libs for `net/http` would've been an alternative solution and would've worked out of the box. This approach, however, gives us a bit more lambda-specific information.

pkg/telemetry/otel.go
volmedo: absolutely, great point!

PR #140, opened 2025-02-25
chore: upgrade go-ucanto to v0.3.0

volmedo:
Upgrade `go-ucanto` to the latest version to get the fix for the server not validating invocation audiences (https://github.com/storacha/go-ucanto/pull/38).


alanshaw (approved):
🙏

PR #142, opened 2025-03-11
fix: reduce providers cache entry expiration to 1 day

volmedo:
we are seeing poor hit rates in the providers cache and want to try reducing the expiration to see the effect it has while we investigate the issue.


volmedo:
yup, I agree. Hannah and I were taking a look at the metrics yesterday and we don't really understand why they look the way they look, so we need to investigate a bit more. We were also wondering if the eviction policy was actually what we expected. This is more of an experiment to see what effect it had.

Anyway, I gave it a second look and I'm now inclined to think the issue is that memory size is not enough, so let's try #143 first.

PR #147, opened 2025-03-12
fix: peer ID in legacy synthesized provider result

alanshaw:
Fixes the empty peer ID which should allow the indexer to process records in the queue.

> deserializing message: failed to parse peer ID: invalid cid: cid too short
> https://github.com/storacha/indexing-service/blob/04ac62a233322607e72eb592c11d9b59d742bef8/pkg/aws/sqscachingqueue.go#L120

The hard coded peer ID is actually the peer ID of the storage node we'll be deploying to production shortly.

### DIFF
--- a/pkg/service/providerindex/legacy.go
+++ b/pkg/service/providerindex/legacy.go
@@ -27,8 +27,8 @@ import (
 	"github.com/storacha/go-ucanto/did"
 )
 
-// PeerID is the peer ID used in synthetized provider results.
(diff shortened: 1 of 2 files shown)


volmedo:
could you please remind me what that peer ID is used for? What should it look like when not hardcoded?

alanshaw:
It is used for publishing to IPNI, to tell it the ID of the peer that is providing the data. However we're not publishing this to IPNI, just caching it, so it could be anything really (we could generate a random one each time if we wanted) - it does not appear in query results.

volmedo:
that was my suspicion, thanks for confirming!

volmedo (approved):
LGTM, thanks for taking care of this.

PR #150, opened 2025-03-12
perf: discard caching jobs that will timeout

alanshaw:
1. Prioritize caching the DAG root.
    - This is oftentimes the multihash that'll be queried to obtain the index.
2. Do not hold up the queue on lambda timeout.
    - This is probably a massive DAG that'll exceed lambda timeout. Retrying will cost more writes, probably still not succeed and hold up the fifo queue.

This also adds a graceful shutdown period - [context is canceled before the lambda times out](https://medium.com/@filiplubniewski/how-to-leverage-aws-lambda-timeouts-with-go-context-cancellation-7dacde656540). This allows the cache provider lambda to inspect the error and succeed anyway.

### DIFF
--- a/cmd/lambda/providercache/main.go
+++ b/cmd/lambda/providercache/main.go
@@ -68,7 +68,12 @@ func handleMessage(ctx context.Context, sqsCachingDecoder *aws.SQSCachingDecoder
 	if err != nil {
 		return err
 	}
-	_, err = providerCacher.CacheProviderForIndexRecords(ctx, job.Provider, job.Index)
+	n, err := providerCacher.CacheProviderForIndexRecords(ctx, job.Provider, job.Index)
+	// If we cached 1 or more records then succeed. We don't need to hold up the
+	// queue by re-attempting a cache operation.
+	if n > 0 {
+		return nil
+	}
 	if err != nil {
 		return err
 	}
--- a/cmd/lambda/providercache/main.go
+++ b/cmd/lambda/providercache/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"time"
 
 	"github.com/aws/aws-lambda-go/events"
 	logging "github.com/ipfs/go-log/v2"
@@ -16,6 +17,9 @@ import (
 )
 
 var log = logging.Logger("lambda/providercache")
+
+// gracePeriod is the amount of time we have to clean up before a lambda exits
+var gracePeriod = time.Second * 3
 
 func main() {
 	lambda.Start(makeHandler)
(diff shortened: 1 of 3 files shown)


cmd/lambda/providercache/main.go
```diff
@@ -17,6 +18,9 @@ import (
 
 var log = logging.Logger("lambda/providercache")
 
+// gracePeriod is the amount of time we have to clean up before a lambda exits
+var gracePeriod = time.Second * 3
```
volmedo, line 22: q: why 3 secs?

pkg/service/providercacher/simpleprovidercacher.go
```diff
@@ -18,20 +20,40 @@ func NewSimpleProviderCacher(providerStore types.ProviderStore) ProviderCacher {
 
 func (s *simpleProviderCacher) CacheProviderForIndexRecords(ctx context.Context, provider model.ProviderResult, index blobindex.ShardedDagIndexView) (uint64, error) {
 	written := uint64(0)
+
+	// Prioritize the root
+	rootDigest := link.ToCID(index.Content()).Hash()
```
volmedo, line 25: essentially, adding the root CID to the cache first in case something fails later when adding the shards, right?

pkg/service/providercacher/simpleprovidercacher_test.go
```diff
@@ -34,11 +35,16 @@ func TestSimpleProviderCacher_CacheProviderForIndexRecords(t *testing.T) {
 			shardIndex.SetSlice(shardMhs[i], sliceMhs[i*3+j], blobindex.Position{})
 		}
 	}
+	// the root block should be in the index also
+	shardIndex.SetSlice(shardMhs[0], link.ToCID(testCid1).Hash(), blobindex.Position{})
```
volmedo, line 39: if the root block always needs to be included in the index too, I think `blobindex.NewShardedDagIndexView` should add it itself to the shards map of the new `ShardedDagIndexView`, i.e. do this call in `blobindex.NewShardedDagIndexView`.

pkg/service/providercacher/simpleprovidercacher_test.go
volmedo, line 1: maybe add a test that confirms the root CID gets added even when something fails afterwards. I'm not sure if it adds too much value, so I leave it to your discretion.

cmd/lambda/providercache/main.go
```diff
@@ -69,6 +84,12 @@ func handleMessage(ctx context.Context, sqsCachingDecoder *aws.SQSCachingDecoder
 		return err
 	}
 	_, err = providerCacher.CacheProviderForIndexRecords(ctx, job.Provider, job.Index)
+	// Do not hold up the queue by re-attempting a cache job that times out. It is
+	// probably a big DAG and retrying is unlikely to subsequently succeed.
+	if errors.Is(err, context.DeadlineExceeded) {
+		log.Warnf("not retrying cache provider job for: %s error: %s", job.Index.Content(), err)
+		return nil
```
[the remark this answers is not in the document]
alanshaw: That's the idea yes.

cmd/lambda/providercache/main.go
alanshaw: 3 is less than 5 but more than 1.

pkg/service/providercacher/simpleprovidercacher_test.go
alanshaw: Good point but I think by design the root block doesn't _have_ to be included. Typically it will though and perhaps this should account for that.

cmd/lambda/providercache/main.go
volmedo: XDD

hannahhoward (approved):
LGTM

PR #151, opened 2025-03-13
feat: dev environments share the same cache

### LINKED ISSUE #146: single shared cache for dev environments
Most of the cost of developer deployments of the indexing-service goes to the caches, which AWS will charge for even when they are not used.

Let's update the deployment config so that the different development environments share a single cache deployment. Each environment will still have their own lambdas so that different codes can be tested without interfering with each other.

volmedo:
closes #146 

We want all dev environments to share the same cache, while production and staging keep their dedicated ones as they are today. To do so, this PR:
- keeps the existing config for the caches in the `app` config and replicates the config for the caches in the `shared` config
- as a bonus, the VPC is also shared, which saves also the cost of multiple NATs and other network resources
- config for the caches and VPC in `shared` will only be applied when the environment is not prod or staging
- conversely, config for the caches and VPC in `app` will only be applied if the environment is prod or staging
- to avoid repeating code, config for the caches and VPC was put into a terraform module, with `app` and `shared` instantiating the module as needed
- lambdas are always per-environment, but point to the dedicated or shared cache/VPC as needed

I already deployed the shared VPC and cache for the dev environments, and applied the config to my environment to confirm things are working as expected. I'd like to do a round of cleaning dev envs and I'll ping the team to know if they are using them or not.

A side effect of applying this new config is that the caches and the VPC need to be recreated in prod and staging. This is due to the way terraform keeps track of resource states and the caches/VPC coming from a module now. I don't think this is such a big deal, but will gather opinions from the team. If it is, there is some trick that can be done by manually modifying the state, but I'd rather let tofu do the thing for me.


hannahhoward (approved):
LGTM this is great Vicente.

PR #154, opened 2025-03-18
feat(infra): extra deployments for prod

### LINKED ISSUE #136: Multiple prod deployments
# What

After we remove remaining issues in indexer slow down, we may still be hitting latency of just having a single global indexer. 

We may want to deploy up to 5 global indexers to give ourselves the best TTFB we can get. Cost to run these is relatively low (few hundred a month), especially if we drop the cache size significantly, which currently looks pretty doable.

Let's hold on doing this till all other optimizations ship, but this may be the next required step.

We'd need to setup Route 53 to intelligently route people to the closest indexer, but this is definitely available as a pretty standard Lambda feature

Alternatively, we can consider putting cloud front in front (as we currently do with content claims), but that will likely be more expensive and less like the ultimate network topology

volmedo:
Resolves #136 

This PR updates the indexer's terraform config so that, beyond a main deployment, extra deployments in different regions are done. Extra deployments will only be done for the production environment.

### Implementation details

The main changes are:
- previous configuration is moved to its own `"indexer"` module
- the root module now instantiates a main deployment, and extra deployments if required (thanks to OpenTofu 1.9.0, [provider blocks can be instantiated dynamically](https://opentofu.org/blog/opentofu-1-9-0/))
- the split between a main deployment and extra deployments allows not having to change the current region configuration for non-prod environments
- for modules to be compatible with `for_each` they must not declare any provider blocks of their own. Thus, providers for legacy claims resources are extracted from the `indexer` module to the root module
- DNS records for the API gateway are upgraded to latency-based records. Route53 will route requests to the deployment with the least latency from the request's origin


alanshaw:
I'm curious how this works with storage nodes submitting location commitments for the indexer to cache?

volmedo:
the three deployments are independent, so the storage node will get redirected to the closest (in terms of latency) indexer and the location commitment will be cached there. If a client needing that commitment is closer to a different indexer, the thing won't be cached there, obviously.

deploy/app/modules/indexer/gateway.tf
```diff
@@ -142,4 +142,10 @@ resource "aws_route53_record" "api_gateway" {
     zone_id                = aws_apigatewayv2_domain_name.custom_domain.domain_name_configuration[0].hosted_zone_id
     evaluate_target_health = false
   }
-}
\ No newline at end of file
+
```
hannahhoward, line 145: is this all you need to do to get proper routing by region? That's kind of awesome.

hannahhoward (commented):
This is excellent @volmedo 

Have you managed to test it (other than the modularized indexer)? Do you have an idea of how you can test it, other than deploying to prod (I guess that's ok)

hannahhoward (approved):

deploy/app/modules/indexer/gateway.tf
volmedo: right? From what I read, they take periodic latency measurements and approximate where the request is coming from. I'm sure it won't be perfect, but I'm looking forward to seeing it in action.

volmedo:
I did a deployment to my env to confirm all the resources were created as expected. I also thought of testing it by using a VPN to generate requests from different places, but it looked easier to just test it in prod 😄

deploy/app/modules/indexer/gateway.tf
hannahhoward: yea makes sense. I have an additional proposal after that for getting to optimum performance

PR #157, opened 2025-03-19
fix: report failures so successfully cached provider info is not retried

alanshaw:
Return the failed items on error so that the successful items are not retried.

### DIFF
--- a/cmd/lambda/providercache/main.go
+++ b/cmd/lambda/providercache/main.go
@@ -25,6 +25,11 @@ func main() {
 	lambda.Start(makeHandler)
 }
 
+type handlerResult struct {
+	id  string
+	err error
+}
+
 func makeHandler(cfg aws.Config) any {
 	providersRedis := goredis.NewClient(&cfg.ProvidersRedis)
 	if cfg.HoneycombAPIKey != "" {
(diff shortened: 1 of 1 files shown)


cmd/lambda/providercache/main.go
```diff
@@ -46,6 +46,9 @@ func makeHandler(cfg aws.Config) any {
 			}
 		}
 
+		batchItemFailures := []events.SQSBatchItemFailure{}
+		var failureMutex sync.Mutex
```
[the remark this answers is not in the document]
alanshaw: good call

cmd/lambda/providercache/main.go
```diff
@@ -25,6 +25,11 @@ func main() {
 	lambda.Start(makeHandler)
 }
 
+type handlerResult struct {
+	id  string
+	err error
+}
```
volmedo, line 31: no need to change it, just commenting in case you are curious: it is common practice to declare these types that are scoped to a function within the function itself, right before creating the channel

volmedo (approved):

cmd/lambda/providercache/main.go
alanshaw: FML

PR #163, opened 2025-03-20
feat(infra): self-defined cache cluster with LFU eviction policy

volmedo:
Change our current valkey serverless cluster into a valkey cluster with cluster mode enabled and update service redis clients to the new architecture.

### Cluster mode
There are two types of Elasticache clusters: with cluster mode disabled (CMD) and with cluster mode enabled (CME). CMD clusters are essentially a primary node with some optional replicas for high availability. Only one of these node groups is possible with CMD. CME, of course, allows several node groups (i.e. shards) and will do sharding of data slots between groups.

In our use case, since we are only defining a single shard, both are similar. The main difference is that, when cluster mode is enabled, the cluster offers a configuration endpoint that can be used by the client to automatically discover nodes to route requests to. The `go-redis` library we use handles transparently what node to send requests to when in cluster mode. With CMD you get a primary endpoint and a reader endpoint and we'd need to decide ourselves where to send each type of query.

I went with CME because it makes client logic a bit more straightforward, but using CMD shouldn't be difficult. I'll keep an eye on the metrics to see if using CME comes with a performance penalty.

### Forcing deployments to use a Redis/Valkey cluster
I changed all redis clients to ClusterClients, which forces standalone deployments to use a cluster too. This is not ideal if we want to see users run their own indexers, but the reality is that nobody is doing that today. It can probably be solved by using `redis.UniversalClient`, but we can do it in the future when this is an actual problem.

## Tests
I did some manual testing and the traces confirmed that using `ReadOnly: true` and `RouteRandomly: true` makes the service send requests to the three nodes in the cluster.

## Note for reviewers
This PR is based on my previous PR for multi-region deployments #154. I'll rebase on top of main once the other one is merged.


deploy/app/modules/indexer/elasticcache.tf
```diff
@@ -10,11 +10,10 @@ module "caches" {
...
   environment = terraform.workspace
-  
-  cache_limits = {
-    data_storage_GB = terraform.workspace == "prod" ? 20 : 1
-    ecpu_per_second = terraform.workspace == "prod" ? 10000 : 1000
-  }
+
+  # cache.m7g.large => 2 vCPUs, 6.38 GiB memory, 12.5 Gigabit network, $0.1264/hour
+  # cache.t4g.micro => 2 vCPUs, 0.5 GiB memory, 5 Gigabit network, $0.0128/hour
+  node_type = terraform.workspace == "prod" ? "cache.m7g.large" : "cache.t4g.micro"
```
alanshaw, line 16: Just noting this is a lot less than 20GB...

deploy/modules/elasticaches/main.tf
```diff
@@ -16,33 +16,56 @@ resource "aws_kms_key" "cache_key" {
...
+  port                 = 6379
+  parameter_group_name = aws_elasticache_parameter_group.cluster_enabled.name
 
-  daily_snapshot_time  = "02:00"
-  description          = "${var.environment} ${var.app} ${each.key} cache serverless cluster"
-  kms_key_id           = aws_kms_key.cache_key.arn
-  major_engine_version = "7"
-  security_group_ids   = [aws_security_group.cache_security_group.id]
+  num_node_groups         = 1 # 1 shard
+  replicas_per_node_group = 2 # 2 read replicas per shard
```
alanshaw, line 32: Are the read replicas necessary?

deploy/app/modules/indexer/elasticcache.tf
volmedo: according to our current metrics, it looks like we are able to store more than 2M items with the 20GB. Hannah mentioned that 1M items should probably cover 50% of requests, so I wanted to try with something smaller to see how it goes.

deploy/modules/elasticaches/main.tf
volmedo: I think some redundancy in case a node fails is desirable, and 3 nodes are more than 1 but less than 5 😄

volmedo: btw, the current serverless cluster already replicates data in 3 AZs for redundancy

hannahhoward: everyone says 3 nodes is the magic number

deploy/modules/elasticaches/main.tf
```diff
@@ -16,33 +16,56 @@ resource "aws_kms_key" "cache_key" {
...
+  user_group_ids             = [aws_elasticache_user_group.cache_user_group.user_group_id]
+
+  apply_immediately = true
+}
+
+resource "aws_elasticache_parameter_group" "cluster_enabled" {
+  name   = "${var.environment}-${var.app}-valkey8-params"
+  family = "valkey8"
+
+  parameter {
```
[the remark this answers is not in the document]
volmedo: I don't think having more than a single shard is a requirement to use cluster mode enabled. I agree replication can also be achieved with cluster mode disabled. As I mentioned in the PR description, I chose CME because the routing of the requests is handled by the cluster, making it transparent to the application.

If we disable cluster mode we'll need to have different clients for reads and writes. It's not a big deal, but I don't see what the drawbacks are of using CME.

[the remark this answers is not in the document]
hannahhoward: well the main one is multiple databases, which is disabled in cluster mode, and kind of what we'd need for my data tiering idea (since they start so expensive and large) but honestly, everyone says multiple databases in redis is fundamentally a terrible feature and no one should use it. so maybe no on my data tiering plans :P

deploy/app/modules/indexer/elasticcache.tf
volmedo: ok, let's get this merged and we can iterate afterwards. I'll change the instance type, though "practically same price" is ~40% more :P

PR #166, opened 2025-03-21
feat(providerindex): cache empty provider results

hannahhoward:
# Goals

Handle 404's quicker for frequently requested but not found content.
Reason it matters:
- absent caching of 404's, we make a roundtrip to the indexer & cloudflare R2-- this is much slower than anything we do in content claims, causing the total round trip to take longer
- this leads to heavy concurrency usage on Lambda

# Implementation

- makes a seperate cache to record lack of results for a given multihash
- expiration times are short (30 min)
- presence of results takes precedence over previous no results found that is cached -- matters cause it solves invalidation for new content (cache / publish will write to the cache)
- deletion is just handled via expiration.


pkg/service/providerindex/providerindex.go
```diff
@@ -175,6 +186,17 @@ func (pi *ProviderIndexService) cacheResults(ctx context.Context, s trace.Span,
 	return results, nil
 }
 
+// Helper function to cache empty results.
+func (pi *ProviderIndexService) cacheNoProviderResults(ctx context.Context, s trace.Span, mh mh.Multihash) ([]model.ProviderResult, error) {
+	s.AddEvent("caching results")
```
alanshaw, line 191:
```suggestion
	s.AddEvent("caching no results")
```

alanshaw (approved):
LGTM

pkg/service/providerindex/providerindex.go
```diff
@@ -175,6 +186,17 @@ func (pi *ProviderIndexService) cacheResults(ctx context.Context, s trace.Span,
...
+	err := pi.noProviderStore.Set(ctx, mh, struct{}{}, true)
```
volmedo, line 192: for some reason, this is setting the expiration time to the 1 hour default rather than the configured time. As it is today, you need to do a `Set` and then a `SetExpirable`.

If nobody remembers why this is this way, IMO we should just change `Set` so that it sets the configured expiration time when setting the key. Can be done in a different PR.

pkg/service/providerindex/providerindex.go
```diff
@@ -155,6 +163,9 @@ func (pi *ProviderIndexService) getProviderResults(ctx context.Context, mh mh.Mu
 	if legacyRes.err != nil {
 		queryError = errors.Join(queryError, fmt.Errorf("fetching from legacy services failed: %w", legacyRes.err))
 	}
+	if queryError == nil {
+		return pi.cacheNoProviderResults(ctx, s, mh)
+	}
```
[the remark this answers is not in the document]
hannahhoward: I disagree on this one. above are two gaurd clauses and comment making it clear we're not in the "no results" section of the code.

we shouldn't be putting in unneccesary logic to make code readable -- it's good to name things well, take all steps to keep things readable, etc. But logic is logic, and adding/changing logic is not a good way to document code.  If there's a need for additional clarity, add more comments.

pkg/service/providerindex/providerindex.go
hannahhoward: wow no there's a bunch. including a fix -- thanks for catching that.

pkg/service/providerindex/providerindex.go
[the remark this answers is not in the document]
volmedo: oh, I missed the comment above somehow

pkg/aws/service.go
```diff
@@ -157,6 +158,13 @@ func FromEnv(ctx context.Context) Config {
 					MinVersion: tls.VersionTLS12,
 				},
 			},
+			NoProviderRedis: goredis.ClusterOptions{
+				Addrs:                      []string{mustGetEnv("NO_PROVIDERS_REDIS_URL")},
+				CredentialsProviderContext: redisCredentialVerifier(awsConfig, mustGetEnv("REDIS_USER_ID"), mustGetEnv("PROVIDERS_REDIS_CACHE")),
```
alanshaw, line 163:
```suggestion
				CredentialsProviderContext: redisCredentialVerifier(awsConfig, mustGetEnv("REDIS_USER_ID"), mustGetEnv("NO_PROVIDERS_REDIS_CACHE")),
```

alanshaw (approved):
LGTM, although I've added 1 change that I think _must_ be applied.

pkg/aws/service.go
volmedo: great catch, thanks!

PR #172, opened 2025-03-21
perf: increase batch size

alanshaw:
Attempt to consume the queue faster.

Screenshot 2025-03-21 at 15 41 02


volmedo (commented):
I think this is not the change you want to do to increase the batch size. Looking at [this doc](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html) it looks like the `maxReceiveCount` in a redrive policy is how many times processing a message needs to fail before the message is sent to the DLQ.

This is the line you were looking for I believe: https://github.com/storacha/indexing-service/blob/95b1caa534bb517d5a6fc36ffb006a410393593a/deploy/app/modules/indexer/lambda.tf#L365

alanshaw:
Oops my bad!

PR #173, opened 2025-03-24
perf: concurrent provider cache

alanshaw:
This PR alters the provider cacher code to concurrently cache provider records up to 5 at a time.

In manual testing with an atrificial network delay I was able to see caching 10k provider records reduce from 20s to around 5s.

The aim is to process the caching queue faster as it is continually growing atm.


volmedo (approved):
LGTM! 🚀

PR #180, opened 2025-03-26
chore(dns): alternate domain for API gateway, main domain for cloudfront

volmedo:
For prod, we want the main `indexer.storacha.network` domain to point to the cloudfront distribution. The current config uses that domain name both as a custom domain for API gateway and as an alternate domain for the cloudfront distribution.

Instead of that, we will give the API gateway its own `api.indexer.storacha.network` domain to make it clearer which resource is using which domain name. This allows removing references to the gateway's custom domain in cloudfront route53 records, and references to the cloudfront domain in the gateway's config.

### DIFF
--- a/deploy/app/cloudfront.tf
+++ b/deploy/app/cloudfront.tf
@@ -1,6 +1,7 @@
 locals {
   # Only prod gets a CloudFront distribution
   should_create_cloudfront = terraform.workspace == "prod"
+  cloudfront_alternate_domain = "${var.app}.storacha.network"
 }
 
 resource "aws_cloudfront_distribution" "indexer" {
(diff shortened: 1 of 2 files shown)


alanshaw (approved):
LGTM. Why do we not have cloudfront for staging? Would be good to verify cloudfront config works in staging before promoting to production. Is there running cost when not used?

volmedo:
I've been reviewing CloudFront pricing and it looks they charge by data transfer, HTTP requests and edge computation, so it should be ok to enable in staging. I agree is desirable for staging to be as similar as production as possible, especially if it doesn't imply a huge investment.

I'll open a PR for that.

volmedo:
https://github.com/storacha/indexing-service/pull/183

PR #188, opened 2025-03-28
perf(lambda): up memory to 1GB for prod

hannahhoward:
# Goals

Speed up lambda executions -- we're giving the indexer currently only 1/8 the memory we give content claims. Importantly, CPU in Lambda is a function of memory, meaning we give it less than 1/8 the compute power.

# Implementation

Up to 1GB for prod like content claims


volmedo (approved):
🚀

PR #189, opened 2025-03-29
fix(telemetry): introduce sample ratio

hannahhoward:
# Goals

We're once again overconsuming our capacity on Honeycomb, cause while we're limiting spans to those with a parent on API calls, now that the provider caching job is working, we're recording tons of child span interactions with redis
Screenshot 2025-03-28 at 5 57 11 PM

# Implementation

for spans that have no parent, introduce a sampling ratio, to prevent overloading honeycomb.


travis (approved):
I don't Go and I haven't used honeycomb, but the idea makes sense and the code generally seems like what I'd expect, :shipit:

PR #193, opened 2025-04-08
fix: increase timeout to 5m for POST /claims

alanshaw:
Tentatively increases the timeout for the `POST /claims` handler to 5 minutes. Invoking `assert/index` with indexes that have 10k+ items can take a long time to cache.

I'm not sure we really want to wait for all this...


hannahhoward (approved):
LGTM

PR #202, opened 2025-05-08
fix: increase cache queue concurrency

alanshaw:
re: https://github.com/storacha/indexing-service/pull/200#pullrequestreview-2822538569

Goroutines are cheap, let see if simply increasing job concurrency will be fast enough.


volmedo (approved):

volmedo:
Given that the bottleneck is likely the RTTs to the redis cache, and that redis is single-threaded, I still think batching commands in a redis pipeline would get us better results.

But let's try this.

alanshaw:
yep, that's my next goto

PR #203, opened 2025-05-08
feat: add `GET /.well-known/did.json` route

alanshaw:
Serve a DID document like the following at `/.well-known/did.json`. It allows external entities to resolve and verify our `did:web`:

```json
{
  "@context": "https://w3id.org/did/v1",
  "id": "did:web:indexer.storacha.network",
  "verificationMethod": [
    {
      "id": "did:web:indexer.storacha.network#owner",
      "type": "Ed25519VerificationKey2020",
      "controller": "did:web:indexer.storacha.network",
      "publicKeyMultibase": "z6MknTPdpBGfFysJWCJn7SbhrAA7Xmp6ZCnjH9dPxJy2T7sk"
    }
  ],
  "assertionMethod": [
    "did:web:indexer.storacha.network#owner"
  ],
  "authentication": [
    "did:web:indexer.storacha.network#owner"
  ]
}
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -235,8 +235,6 @@ func GetClaimsHandler(service types.Querier) http.HandlerFunc {
 
 // Document is a did document that describes a did subject.
 // See https://www.w3.org/TR/did-core/#dfn-did-documents.
-// TODO: Should we include the non-standard publicKey field for compatability reasons?
-// See https://w3c.github.io/did-spec-registries/#publickey
 type Document struct {
 	Context            []string             `json:"@context"` // https://w3id.org/did/v1
 	ID                 string               `json:"id"`
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -89,6 +90,7 @@ func NewServer(indexer types.Service, opts ...Option) *http.ServeMux {
 	mux.HandleFunc("GET /claim/{claim}", GetClaimHandler(indexer))
 	mux.HandleFunc("POST /claims", PostClaimsHandler(c.id, indexer, c.contentClaimsOptions...))
 	mux.HandleFunc("GET /claims", GetClaimsHandler(indexer))
+	mux.HandleFunc("GET /.well-known/did.json", GetDIDDocument(c.id))
 	return mux
 }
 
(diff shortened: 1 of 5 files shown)


pkg/server/server.go
```diff
@@ -230,3 +232,54 @@ func GetClaimsHandler(service types.Querier) http.HandlerFunc {
...
+			{
+				ID:                 vid,
+				Type:               "Ed25519VerificationKey2020",
+				Controller:         id.DID().String(),
+				PublicKeyMultibase: strings.TrimPrefix(s.Unwrap().DID().String(), "did:key:"),
+			},
+		}
+		doc.Authentication = []string{vid}
+		doc.AssertionMethod = []string{vid}
+	}
```
volmedo, line 275: nit: maybe always use `s` rather than `id` in the `if`'s body since it has already been type-asserted?
```suggestion
	if s, ok := id.(signer.WrappedSigner); ok {
		vid := fmt.Sprintf("%s#owner", s.DID())
		doc.VerificationMethod = []VerificationMethod{
			{
				ID:                 vid,
				Type:               "Ed25519VerificationKey2020",
				Controller:         s.DID().String(),
				PublicKeyMultibase: strings.TrimPrefix(s.Unwrap().DID().String(), "did:key:"),
			},
		}
		doc.Authentication = []string{vid}
		doc.AssertionMethod = []string{vid}
	}
```

pkg/server/server_test.go
```diff
@@ -290,3 +291,47 @@ func TestGetClaimsHandler(t *testing.T) {
...
 }
+
+func TestGetDIDDocumentHandler(t *testing.T) {
+	t.Run("success", func(t *testing.T) {
+		svr := httptest.NewServer(GetDIDDocument(testutil.Service))
+		defer svr.Close()
+
+		res, err := http.Get(svr.URL)
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, res.StatusCode)
```
volmedo, line 302: OCD: please add a `defer res.Body.Close()` (also in the sub-test below)

pkg/server/server_test.go
```diff
@@ -290,3 +291,47 @@ func TestGetClaimsHandler(t *testing.T) {
...
+
+		bytes, err := io.ReadAll(res.Body)
+		require.NoError(t, err)
+
+		var doc Document
+		err = json.Unmarshal(bytes, &doc)
+		require.NoError(t, err)
+
+		require.Equal(t, doc.ID, didweb.DID().String())
```
volmedo, line 335: what about checking also some field specific of the case where a `WrappedSigner` is used?

PR #205, opened 2025-05-21
feat: configure IPNI endpoint and announce URLs

alanshaw:
Allows the IPNI endpoint for queries to be configured so that when we have deployed our own IPNI instance we can use it for querying.

Allows the IPNI announce URLs to be configured so that when we have deployed our own IPNI instance we can announce to it as well as cid.contact.

### DIFF
--- a/pkg/aws/service.go
+++ b/pkg/aws/service.go
@@ -78,6 +78,7 @@ type Config struct {
 	MetadataTableName                 string
 	IPNIStoreBucket                   string
 	IPNIStorePrefix                   string
+	IPNIAnnounceURLs                  []url.URL
 	NotifierHeadBucket                string
 	NotifierTopicArn                  string
 	ClaimStoreBucket                  string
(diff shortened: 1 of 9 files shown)


pkg/aws/service.go
```diff
@@ -197,8 +208,9 @@ func FromEnv(ctx context.Context) Config {
 					MinVersion: tls.VersionTLS12,
 				},
 			},
-			IndexerURL:             mustGetEnv("IPNI_ENDPOINT"),
-			PublisherAnnounceAddrs: []string{ipniPublisherAnnounceAddress},
+			IndexerURL:                  mustGetEnv("IPNI_ENDPOINT"),
+			PublisherAnnounceAddrs:      []string{ipniPublisherAnnounceAddress},
+			PublisherDirectAnnounceURLs: ipniPublisherDirectAnnounceURLs,
```
volmedo, line 213: would you mind renaming these fields so that it is clear what they are? Probably prefixing them with `IPNI` would be enough.

pkg/presets/presets.go
```diff
@@ -1,5 +1,7 @@
 package presets
 
+var IPNIAnnounceURLs = []string{"https://cid.contact/announce"}
```
volmedo, line 3: I think we could also have a preset for the IPNI URL and make the `IPNI_ENDPOINT` env var not required.

volmedo (approved):
LGTM, I left a couple minor comments, nothing blocking.

volmedo (approved):
thanks for the renames! 💯

PR #221, opened 2025-06-20
fix: content CID is not a shard CID

alanshaw:
While debugging the indexing service cache I found this error. I'm not 100% certain I understand what this field is for in the metadata but it is not for the DAG root CID.


hannahhoward (approved):
This is correct. It should be blank.

PR #222, opened 2025-06-20
fix: limit results from legacy block index table

alanshaw:
Some blocks are in MANY MANY CARs - sometimes because it's a popular block (`LICENCE.md` for example) and other times we had a bug that meant that the same content got uploaded many times over in different CAR files.

This PR adds a limit of 25 to the legacy block index table query results. The same value that is used in the legacy content claims service.

It prevents the indexing service from using too much cache space for a given hash and ensures quick response times in cases where a block is found in many CARs.

### DIFF
--- a/pkg/aws/dynamoblockindextable.go
+++ b/pkg/aws/dynamoblockindextable.go
@@ -13,6 +13,12 @@ import (
 	"github.com/storacha/indexing-service/pkg/types"
 )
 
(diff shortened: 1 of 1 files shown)


hannahhoward (approved):
+1 Approved

is 25 too high anyway?

alanshaw:
> is 25 too high anyway?

🤷 pick a number? 5?

Issue #225, opened 2025-06-23
Post-deployment-strategy-change cleanup (code)


volmedo:
Remove lambda code and refactor service initialization now that supporting both deployment strategies is not needed anymore.

PR #226, opened 2025-06-23
feat: containerize for AWS

### LINKED ISSUE #223: Containerize indexing-service
The indexing-service is deployed as a set of lambdas. Most of them are triggered by API gateway endpoints, while the rest are launched in response to periodic, cron-like events, SNS notifications or SQS messages.

To allow running on ECS, the indexing-service must be deployed as a monolithic, containerized service. Usage of services such as S3 and DynamoDB will be kept, but others, like API gateway and SNS will not be necessary. For instance, API gateway can be replaced by a standard HTTP server now, and standard inter-process communication mechanisms can be used instead of SNS.

volmedo:
## Why

Closes #223

We want to move the indexing-service from the current lambda-based deployment to a fargate task running in ECS. The first step to achieve this is merging the different functions done by the lambdas into a single, containerized service.

## How

The indexing-service features are implemented via the following lambdas:
- `getclaim`
- `getclaims`
- `getdiddocument`
- `getroot`
- `postclaims`
- `notifier`
- `remotesync`
- `providercache`

### HTTP endpoint lambdas

`getclaim`, `getclaims`, `getdiddocument`, `getroot` and `postclaims` are all HTTP lambdas triggered by requests to API Gateway endpoints. These can be easily replaced by a standard HTTP server exposing the different routes.
https://github.com/storacha/indexing-service/blob/e91e53712af9a5470cbe321418055be6507037db/cmd/aws.go#L83

### `notifier` and `remotesync`

These two lambdas work in tandem. `notifier` fetches the latest advert from IPNI and compares that with the current chain. For adverts that had already been ingested by IPNI, it notifies the `remotesync` lambda, that will set the corresponding provider info entries as expirable in the cache. It's ok for these entries to be evicted from the cache now that they can be found in IPNI.

`notifier` is triggered by a periodic CloudWatch event, which is replaced by a regular ticker. It triggers the `remotesync` lambda by issuing a SNS notification. Since both components will be running in the same service now, they can be just connected together. The `setupIPNIPipeline` function creates a `Notifier` and a `RemoteSyncer` and sets them up so that the syncer is invoked directly when needed.
https://github.com/storacha/indexing-service/blob/e91e53712af9a5470cbe321418055be6507037db/cmd/aws.go#L87-L109

### `providercache`

The main goal of the indexing-service is to cache content claims before they are available through IPNI. Caching index claims can take a long time for big indexes, as an entry is stored for each shard. Therefore, this is done asynchronously.

When the indexing service fetches an index, it will store it in an S3 bucket and put a message in a SQS queue with the corresponding details. The `providercache` lambda will be triggered by messages in the SQS caching queue, and do the actual store in the cache.

Given that the indexing-service will run as a monolithic service, the SQS queue could be replaced with some local queue. Using the SQLite-backed job queue used in Piri would be an option, since we want the queue to be persistent and survive service restarts. I kept the SQS queue for now to simplify this PR as much as possible. This also allows the queue to be shared by multiple tasks when the service is scaled horizontally, sharing the load to cache indexes. But we can consider getting rid of the SQS queue if an alternative would be more cost-efficient.

SQS queues have the ability to trigger a lambda. They can also trigger ECS tasks by defining an EventBridge Pipe, but I'm not sure whether the pipe can target an already running task or it spawns its own. Again, to simplify the changes in this PR I created a simple poller that fetches batches of messages periodically from the queue and caches them. This approach can be revisited in the future if we prefer a push-based mechanism.

As part of the implementation of this new component, I moved some interface definitions from the `blobindexlookup` package to the `providercacher` package and added new methods to the `CachingQueue` interface so that caching jobs can also be read and deleted from the queue.

## Known issues

- Service initialization is a bit weird right now because I didn't touch the function that builds the service for the lambdas, as that needs to be working until the deployment configuration is updated. Don't pay too much attention to that since it will be taken care of afterwards (see #225).
- The `notifier` lambda was configured to check IPNI [every minute](https://github.com/storacha/indexing-service/blob/e91e53712af9a5470cbe321418055be6507037db/deploy/app/lambda.tf#L388-L392). However, the IPNI notifier has a fixed polling interval of 30 seconds. If this is relevant, let me know and I'll update the IPNI notifier implementation to accept a configurable interval.
- The caching queue poller has a default poll interval of 5 seconds and it will fetch batches of 10 messages. Feel free to suggest alternative values.

## Testing

For now, this needs to be deployed to be tested, so testing will be done in staging once available.

### DIFF
--- a/cmd/aws.go
+++ b/cmd/aws.go
@@ -11,7 +11,7 @@ import (
 	"github.com/storacha/indexing-service/pkg/redis"
 	"github.com/storacha/indexing-service/pkg/server"
 	"github.com/storacha/indexing-service/pkg/service/providercacher"
-	"github.com/storacha/indexing-service/pkg/service/providerindex"
+	"github.com/storacha/indexing-service/pkg/service/providerindex/remotesyncer"
 	"github.com/storacha/indexing-service/pkg/telemetry"
 	"github.com/storacha/ipni-publisher/pkg/notifier"
 	"github.com/storacha/ipni-publisher/pkg/store"
@@ -98,7 +98,7 @@ func setupIPNIPipeline(cfg aws.Config) (*notifier.Notifier, error) {
 	chunkLinksTable := aws.NewDynamoProviderContextTable(cfg.Config, cfg.ChunkLinksTableName)
 	metadataTable := aws.NewDynamoProviderContextTable(cfg.Config, cfg.MetadataTableName)
 	publisherStore := store.NewPublisherStore(ipniStore, chunkLinksTable, metadataTable, store.WithMetadataContext(metadata.MetadataContext))
-	remoteSyncer := providerindex.NewRemoteSyncer(providerStore, publisherStore)
+	remoteSyncer := remotesyncer.New(providerStore, publisherStore)
 
 	// setup notifier to periodically check IPNI and notify remote syncer if updates are required
 	headStore := aws.NewS3Store(cfg.Config, cfg.NotifierHeadBucket, "")


pkg/service/providercacher/cachingqueuepoller.go
```diff
@@ -0,0 +1,163 @@
...
+// processJob processes a single caching job from the queue.
+func (p *CachingQueuePoller) processJob(ctx context.Context, job ProviderCachingJob) error {
+	err := p.cacher.CacheProviderForIndexRecords(ctx, job.Provider, job.Index)
+	if err != nil {
+		// Do not hold up the queue by re-attempting a cache job that times out. It is
+		// probably a big DAG and retrying is unlikely to subsequently succeed.
+		if errors.Is(err, context.DeadlineExceeded) {
+			log.Warnf("not retrying cache provider job for: %s error: %s", job.Index.Content(), err)
+			return nil
+		}
```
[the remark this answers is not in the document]
volmedo: yeah, copy-pasta error. Good catch.

pkg/service/providercacher/cachingqueuepoller.go
```diff
@@ -0,0 +1,163 @@
...
+func WithPollInterval(interval time.Duration) Option {
+	return func(p *CachingQueuePoller) {
+		p.interval = interval
+	}
+}
+
+// WithJobBatchSize sets the maximum number of jobs to process in a batch
+func WithJobBatchSize(size int) Option {
+	return func(p *CachingQueuePoller) {
+		p.maxJobBatchSize = size
```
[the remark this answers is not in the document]
volmedo: turns out this wasn't the max despite its name, but the check is needed because SQS will complain if we try to read a batch of more than 10 messages

pkg/service/providercacher/cachingqueuepoller.go
```diff
@@ -0,0 +1,191 @@
...
+		// Process batch in a new goroutine
+		wg.Add(1)
+		go func(batch []ProviderCachingJob) {
+			defer func() {
+				// Release semaphore when done
+				<-maxGoroutines
+				wg.Done()
+			}()
+
+			p.processBatch(ctx, batch)
```
alanshaw, line 152: I think this will be fine, but you're still restricting the maximum concurrent jobs by waiting for a whole batch to finish. Personally I'd have something reading from the queue and putting individual jobs on a channel, and many channel consumers so you maintain the total number of concurrent tasks.

Right now if you have a slow job in every batch then you'll end up with just `maxParallelBatches` running jobs at times, whereas you really want to maintain `maxParallelBatches * jobBatchSize` at _all_ times.

pkg/service/providercacher/cachingqueuepoller.go
```diff
@@ -0,0 +1,191 @@
...
+				log.Debug("Context canceled, stopping job processing")
+				return
+			default:
+				if err := p.cacher.CacheProviderForIndexRecords(ctx, job.Provider, job.Index); err != nil {
+					log.Errorf("Failed to cache provider %s: %v", job.Provider, err)
+					return
+				}
+
+				// Delete the job if processing was successful
+				if err := p.queue.DeleteJob(ctx, job.ID); err != nil {
```
alanshaw, line 182: We should increase the visibility timeout if the job is taking a long time to cache and/or just set a big visibility timeout initially, assuming we'll be able to cache all the records in say, an hour.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ChangeMessageVisibility.html

pkg/service/providercacher/cachingqueuepoller.go
```diff
@@ -0,0 +1,191 @@
...
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	logging "github.com/ipfs/go-log/v2"
+)
+
+const (
+	defaultPollInterval = 5 * time.Second
```
alanshaw, line 13: Alternatively you could set `WaitTimeSeconds` to the max and not have an interval (or a very short one)...

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-short-and-long-polling.html#sqs-long-polling

alanshaw (approved):

volmedo:
sorry for the back and forth, that was me trying to ship something quickly and not thinking too much. Thank you for keeping me honest here.

I swapped the previous batch processing logic with a jobqueue + long-polling. Now there is a goroutine that long-polls the queue and constantly reads new messages when available. The jobs are then queued in the jobqueue, which will put them in a channel for the workers to process. I set a concurrency of 100 for the queue.

The job processing logic also handles job timeouts properly. A max of 5 minutes is set, which matches the visibility timeout in the queue config and it's what we had before in the lambda. I think that should give plenty of room for records to be cached now that we use redis pipelines. When a job fails because of a timeout, it will be discarded as retrying it is unlikely to make it succeed. However, if the failure is not a timeout, it will be made visible immediately so that it can be picked up again and retried (up to 4 times, after which it will be sent to the DLQ if still unsuccessful).

I think these changes address all your concerns, but please let me know if I'm still missing something.

pkg/service/providercacher/cachingqueue.go
```diff
@@ -8,15 +8,35 @@ import (
...
+		Read(ctx context.Context, maxJobs int) ([]ProviderCachingJob, error)
+	}
+
+	CachingQueueReleaser interface {
+		Release(ctx context.Context, jobID string) error
+	}
+
+	CachingQueueDeleter interface {
+		Delete(ctx context.Context, jobID string) error
+	}
```
alanshaw, line 25: I would argue that the reader interface should include release and delete functions...it is maybe a "consumer".

pkg/service/providercacher/cachingqueuepoller.go
```diff
@@ -0,0 +1,180 @@
...
+func (p *CachingQueuePoller) processJobs(ctx context.Context) {
+	// Read a batch of jobs and queue them in the job queue
+	jobs, err := p.queue.Read(ctx, p.jobBatchSize)
+	if err != nil {
+		log.Errorf("Error reading jobs from queue: %v", err)
+		return
+	}
+
+	for _, job := range jobs {
+		err := p.jq.Queue(ctx, job)
```
alanshaw, line 142: Is there any backpressure here? Are we going to buffer the entire queue in memory? I _think_ this blocks if the concurrency limit is reached but worth checking.

alanshaw (approved):

pkg/service/providercacher/cachingqueue.go
volmedo: the convention in go is adding an "er" suffix to one-method interfaces. I find that a bit confusing, to be honest, because the "reader" will be the one calling the "read" method, and not the one implementing it, which would be a "readable". Therefore, if I were to put all these together in a single interface I would call it "consumable" instead. More probably, I'd just collapse all the one-method interfaces in the single `CachingQueue` interface and call it a day.

I'll leave it as is for now if you don't mind. I don't think it's hard to read and understand as it is and we will likely not need to touch it too much in the future anyway.

pkg/service/providercacher/cachingqueuepoller.go
volmedo: yes, you are right, it is limited by the queue's concurrency.

`jobqueue.Queue` sends the job to the channel where the job queue workers read from. I don't use the `WithBuffer` option when creating the queue, so this channel is unbuffered and will block as soon as there are no workers available to receive new jobs from the channel.

Issue #216, opened 2025-06-16
Do not fail queries when caching fails


hannahhoward:
Implement https://github.com/storacha/RFC/pull/57

PR #228, opened 2025-06-23
fix: do not fail queries when caching fails

alanshaw:
* If caching fails simply log the error.
* Configue Sentry in project so errors can be alerted on.
* Expose sentry as a `ipfs/go-log` logger - calls to `log.Errorf()` etc. are automatically sent to Sentry.
* `ProviderIndex` now takes options (for configuring the logger) so I also removed the `NewWithClock(...)` constructor and added a `WithClock(...)` option.

Note: only the providerindex uses the `SentryLogger` right now.

resolves https://github.com/storacha/indexing-service/issues/216


volmedo (approved):
Love the refactors. I left some non-blocking comments.

That said, I think I'd keep logging separated from error reporting and just add a function to the telemetry package to report the exception to Sentry if configured. It would require adding calls all over the place, but they would be explicit, which I think is an advantage. It would also allow choosing what to report and what not to.

Just my opinion. I'm approving, feel free to disregard.

Issue #218, opened 2025-06-16
Setup tracing for publish and cache


hannahhoward:
Possibly already done?

Insure that we have solid tracing on what takes a long time in any given cache or publish operation on the indexer

PR #231, opened 2025-06-26
feat: add tracing to publish and cache operations

alanshaw:
Adds attributes to publish tracing and adds tracing and attributes to caching ops.

resolves https://github.com/storacha/indexing-service/issues/218


volmedo (approved):
LGTM.

I think adding some calls to `telemetry.Error` when something goes wrong would be useful for debugging (maybe not all errors are interesting enough to be reported though).

frrist (approved):

PR #234, opened 2025-06-30
fix: hot and warm network route53 zone conflicts

volmedo:
Adding different networks introduces different "levels of sharedness" to shared resources. Some of them are truly shared amongst all envs no matter what network they serve, while others are shared by envs within the same network.

This specifically applies to the Route53 hosted zone. I thought a single resource could be shared but that's not how it works. This PR adds another zone for the warm network.

I checked the plans in CI and things look good. No unexpected changes to prod or staging, but certs and validations records fixed for warm-staging.


alanshaw (approved):
❤️ thank you

Issue #213, opened 2025-06-11
Turning off old.web3.storage retrievals via the indexer service


hannahhoward:
# What

We want to disable retrievals of old.web3.storage content that hasn't been migrated to Storacha by June 13. The purpose of doing so is that old.web3.storage content is supposed to be deleted started on July 13th, and we're using this to make sure anyone who needs their data but has managed to miss the many deprecation emails we've sent will learn of the issue when their content isn't served.

The complicating factor is that all content (nft.storage, old.web3.storage, web3.storage, storacha.network) is fundamentally co-mingled together in the same bucket. Moreover, we don't want to disable retrievals for those users that migrate to Storacha, which doesn't actually move their data but instead just invokes invocations to record them as part of the new Storacha network.

# Proposed approach

Currently, content claims for old.web3.storage content is fetched are materialized in real time using data from the block index table. This applies also to new data. However, the key pattern used for old.web3.storage is different. Specifically, it is of the form `{region}/{bucket}/{key}` and specifically should begin `us-west-2/dotstorage-prod-1`. 

There's already a code path to handle this different key path right here:
https://github.com/storacha/indexing-service/blob/main/pkg/aws/blockindextablemapper.go#L80

By instead simply ignoring these keys and returning no results we in effect turn off retrievals for old.web3.storage. However, there are two additional caveats:
- this would turn off nft.storage retrievals as well if we filter all of these, so we need to only ignore those that begin `us-west-2/dotstorage-prod-1`
- because we don't change keys when we migrate old web3 storage users to storacha, we need an additional check to make sure we continue to serve this content. the proposed approach is that once we have the shard itself, we connect the indexer to two dynamo tables: blob-registry and the store table. because the migration step for moving to storacha.network involves making store/add or blob/add invocations for these shards without actually uploading, there should be an entry in one of these two tables for the shard itself.

So, our algorithm is:
- for each result in block index table
   - if result is new key format, return the entry
   - if result is old key format for nft.storage, return the entry (we can use this same approach for nft.storage later)
   - if result is old key format for web3.storage
      - AND entry for shard exists in blob-registry or store table, return the entry
      - otherwise, ignore this entry

# Caveats

We expect there is a possibility there is something we haven't considered. If:
1. We discover our approach blocks most but not all old.web3.storage content, we should go ahead and ship and then keep improving
2. We discover our approach blocks content that should not be blocked, then we should NOT ship it

[...]

PR #235, opened 2025-07-09
feat(aws): skip unmigrated dotstorage shards

hannahhoward:
# Goals

implements blocking of retrievals by skipping old web3.storage cids that are not listed in the store table or the blob-registry table per #213 

# Implementation

- in block index table mapper
   - skip shards if they are dot storage shards and are not listed as migrated
- implement migrated shard checker
   - checks both store table and blob registry table to see if a shard is migrated
- update deploy configs

# For Discussion

Before we ship this, I really need to test round trip by deploying an indexer env connected to prod

### DIFF
--- a/deploy/app/legacyclaims.tf
+++ b/deploy/app/legacyclaims.tf
@@ -30,6 +30,31 @@ variable "legacy_block_index_table_region" {
 }
 
 
+variable "legacy_store_table_name" {
+  description = "The name of the legacy store DynamoDB table"
+  type = string
+  default = ""
+}
+
+# the store table is always deployed in us-west-2 for both prod and staging
+variable "legacy_store_table_region" {
+  description = "The region where the legacy store DynamoDB table is provisioned"
+  type = string
+  default = "us-west-2"
+}
+
+variable "legacy_blob_registry_table_name" {
+  description = "The name of the legacy blob registry table DynamoDB table"
+  type = string
+  default = ""
+}
+
+# the blob registry table is always deployed in us-west-2 for both prod and staging
+variable "legacy_blob_registry_table_region" {
+  description = "The region where the legacy blob registry table DynamoDB table is provisioned"
+  type = string
+  default = "us-west-2"
+}
 
 variable "legacy_allocations_table_name" {
   description = "The name of the legacy w3infra allocation DynamoDB table"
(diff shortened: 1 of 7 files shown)


alanshaw:
I think there's a little confusion here but should be easily fixed. The key format for nft.storage is:

```
us-east-1/dotstorage-prod-0/nft-[id]/xxx.car

us-west-2/dotstorage-prod-1/nft-[id]/…
```

For web3:

```
us-east-1/dotstorage-prod-0/[id]/xxx.car

us-west-2/dotstorage-prod-1/[id]/…

```

I.e. no prefix in the key


So there’s 2 things - there are 2 dotstorage buckets (east and west) with suffix 0 and 1. They both have data from nft.storage and web3.storage. If memory serves correctly the east bucket is east-1 but needs to be verified (might be east-2).

deploy/app/legacyclaims.tf
```diff
@@ -30,6 +30,31 @@ variable "legacy_block_index_table_region" {
 }
 
 
+variable "legacy_store_table_name" {
+  description = "The name of the legacy store DynamoDB table"
+  type = string
+  default = ""
+}
+
+# the store table is always deployed in us-west-2 for both prod and staging
```
alanshaw, line 39: Erm, is that right?

deploy/app/legacyclaims.tf
```diff
@@ -30,6 +30,31 @@ variable "legacy_block_index_table_region" {
...
+  default = "us-west-2"
+}
+
+variable "legacy_blob_registry_table_name" {
+  description = "The name of the legacy blob registry table DynamoDB table"
+  type = string
+  default = ""
+}
+
+# the blob registry table is always deployed in us-west-2 for both prod and staging
```
alanshaw, line 52: Ok that's definitely not right!

pkg/aws/blockindextablemapper.go
```diff
@@ -87,6 +93,17 @@ func (bit blockIndexTableMapper) GetClaims(ctx context.Context, contentHash mult
 				continue
 			}
 
+			// check if shard is a web3.storage shard
+			if strings.Join(parts[:2], "/") == "us-west-2/dotstorage-prod-1" {
```
alanshaw, line 97: As per previous comment this needs to be altered to also look for `dotstorage-prod-0` in the east and discount the keys with `nft-` in them.

pkg/aws/dynamomigratedshardchecker.go
```diff
@@ -0,0 +1,89 @@
...
+	return o.Count > 0, nil
+}
+
+func (d *DynamoMigratedShardChecker) ShardMigrated(ctx context.Context, shard ipld.Link) (bool, error) {
+	if migrated, err := d.storeTableShardMigrated(ctx, shard); err != nil {
+		return false, err
+	} else if migrated {
+		return true, nil
+	}
+	return d.blobRegistryTableShardMigrated(ctx, shard)
```
alanshaw, line 79: Note: we'll have to wait for the allocation -> blob registry migration before this will work as desired. I should be able to complete in the next day or two though.

alanshaw (changes requested):

pkg/aws/dynamomigratedshardchecker.go
alanshaw: https://github.com/storacha/project-tracking/issues/270

hannahhoward: i could just look at both tables?

deploy/app/legacyclaims.tf
hannahhoward: ok, my bad, I was just monkey say monkey do copying some other lines.

hannahhoward:
@alanshaw responded to your comments, and did further investigation to confirm the patterns. Unfortunately the prefixed `nft-` is actually two more links down the chain, but seems consistent.

pkg/aws/dynamomigratedshardchecker.go
alanshaw: 🤷 FML sorry I'm so overcommitted right now. If you want to look in both it's fine by me.

alanshaw (approved):
LGTM

PR #238, opened 2025-08-02
refactor(testutil): use external testutil, blobindex

hannahhoward:
# Goals

companion to https://github.com/storacha/go-libstoracha/pull/40 , should not be merged until that PR is merged and in a tagged version

# Implementation

- use external versions go testutil, blobindex, and bytemap
- add testing parameters to all calls to test functions. 
- move extmocks package to pkg/internal root, along with AnyContext function


alanshaw (approved):
LGTM

PR #241, opened 2025-08-11
fix: allow publishing claims by POSTing to `/` too

volmedo:
It looks like piri sends claims as `POST /` instead of `POST /claims`. Let's allow it while it gets fixed.


alanshaw (approved):
❤️ thanks!

PR #242, opened 2025-09-04
fix: migrated shard checker was getting mixed table names

volmedo:
We were interchanging the table names when creating the migrated shard checker, which caused queries to fail.

### DIFF
--- a/pkg/aws/service.go
+++ b/pkg/aws/service.go
@@ -311,10 +311,10 @@ func Construct(cfg Config) (types.Service, error) {
 	blobRegistryTableCfg := cfg.Config.Copy()
 	blobRegistryTableCfg.Region = cfg.LegacyBlobRegistryTableRegion
 	legacyMigratedShardChecker := NewDynamoMigratedShardChecker(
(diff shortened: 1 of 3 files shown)


pkg/aws/service.go
```diff
@@ -311,10 +311,10 @@ func Construct(cfg Config) (types.Service, error) {
 	blobRegistryTableCfg := cfg.Config.Copy()
 	blobRegistryTableCfg.Region = cfg.LegacyBlobRegistryTableRegion
 	legacyMigratedShardChecker := NewDynamoMigratedShardChecker(
-		dynamodb.NewFromConfig(storeTableCfg),
-		dynamodb.NewFromConfig(blobRegistryTableCfg),
 		cfg.LegacyStoreTableName,
+		dynamodb.NewFromConfig(storeTableCfg),
 		cfg.LegacyBlobRegistryTableName,
+		dynamodb.NewFromConfig(blobRegistryTableCfg),
```
volmedo, line 317: here we were giving `LegacyStoreTableName` in the place of `blobRegistryTableName` and viceversa

hannahhoward (approved):

PR #250, opened 2025-10-09
feat: server side support for authorized retrievals

alanshaw:
This PR add server side support for authorized retrievals. This involves 2 different authorized retrieval types:

1. Users can send a delegation for retrieving data from a space (per https://github.com/storacha/indexing-service/pull/249) and it will now be used by the indexer to make a UCAN authorized retrieval (`space/content/retrieve`) from the relevant node (if not cached).
2. When an upload service invokes `assert/index`, the indexer will now obtain a service delegation from the node that is storing the index (using `access/grant`) before making a UCAN authorized retrieval (`blob/retrieve`) to obtain the index. See https://github.com/storacha/RFC/pull/68 for details of on-demand service retrievals.

~~TODO: a test for (2)~~

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -228,7 +228,8 @@ type RetrievalAuth struct {
 	Audience ucan.Principal
 	// Retrieval ability, resource (typically the space) and caveats.
 	Capability ucan.Capability[ucan.CaveatBuilder]
-	// Delegations from the client to the indexing service authorizing retrieval.
+	// Delegations from the client (e.g. `space/content/retrieve`) or a storage
+	// node (e.g. `blob/retrieve`) to the indexing service authorizing retrieval.
 	Proofs []delegation.Proof
 }
 


pkg/service/blobindexlookup/simplelookup.go
```diff
@@ -25,27 +31,98 @@ func NewBlobIndexLookup(httpClient *http.Client) BlobIndexLookup {
...
+		if err != nil {
+			return nil, fmt.Errorf("executing authorized retrieval: %w", err)
+		}
+		defer body.Close()
+		return blobindex.Extract(body)
+	}
+
 	// attempt to fetch the index from provided url
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fetchURL.String(), nil)
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodGet, request.URL.String(), nil)
```
volmedo: I think encapsulating the non-auth retrieval branch in a `doRegularRetrieval` function would make the method more readable.

pkg/types/types.go
```diff
@@ -188,3 +191,63 @@ type Service interface {
...
+//
+// The provided proofs are expected to contain the `space/content/retrieve`
+// delegated capability allowing content to be retrieved using UCAN
+// authorization.
+type RetrievalAuth struct {
+	// The Indexing Service UCAN signing key.
+	Issuer ucan.Signer
+	// Identity of the storage node to retrieve data from.
+	Audience ucan.Principal
+	// Retrieval ability, resource (typially the space) and caveats.
```
volmedo:
```suggestion
	// Retrieval ability, resource (typically the space) and caveats.
```

pkg/types/types.go
```diff
@@ -188,3 +191,63 @@ type Service interface {
...
+	Capability ucan.Capability[ucan.CaveatBuilder]
+	// Delegations from the client to the indexing service authorizing retrieval.
```
volmedo:
```suggestion
	// Delegations from the node to the indexing service authorizing retrieval.
```

volmedo (commented):
this looks good overall. I think the code is becoming a bit unwieldy but and it'd be nice to do a refactor at some point.

to confirm: the indexing-service can be authorized to retrieve (index) blobs by the node that has the blob or directly by the client, yes? In the latter scenario, Alice would issue a delegation to the indexing-service so it can retrieve from a space they have access to, right?

pkg/types/types.go
alanshaw: Ah gotcha. I think "the node" is correct - perhaps "the storage node"

PR #256, opened 2025-10-16
feat(aws): use a publishing queue for IPNI

hannahhoward:
# Goals

update the aws deployment to use an async publishing queue for IPNI publishing rather than trying to publish synchronously

blocked on https://github.com/storacha/go-libstoracha/pull/57

# Implementation

- update dependency for go-libstoracha with shared queue
- update mockery to v3 (cause it doesn't work with go 1.24 otherwise)
- update storoku to v0.4.6
- add ipni publishing queue + bucket
- add ipni publishing queue to aws config using go-libstoracha publishing queue
- update sqs caching queue to use generic version from go-libstoracha
- genericize queue polling struct
- add publishing queue processing to aws command
- add option to completely replace publishing to construct process
- regen mocks
- update caching queue to use generic queue poller
- remove backoff from provider idnex
- add publishqueue poller to use generic poller

### DIFF
--- a/deploy/app/main.tf
+++ b/deploy/app/main.tf
@@ -9,16 +9,16 @@ terraform {
     }
   }
   backend "s3" {
-    bucket = "storacha-terraform-state"
-    key = "storacha/${var.app}/terraform.tfstate"
-    region = "us-west-2"
+    bucket  = "storacha-terraform-state"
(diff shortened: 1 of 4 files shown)


.mockery.yaml
volmedo: out of curiosity, what version of mockery did you use?

volmedo (approved):

PR #288, opened 2026-02-02
feat: support query for equals or location claims

alanshaw:
Will eventually allow Spade to authorize retrievals from Forge. Spade has a list of pieces that make up an aggregate and needs to map from piece CID to blob CID + location URL in order to authorize retrieval. This allows more efficient bulk equivalency mapping of multiple CIDs by not having to consider indexes.

refs https://github.com/storacha/spade-agent/issues/7


fforbeck (approved):

alanshaw:
I'm not sure we need this...there should be no index results for a shard anyway...

